PR #17, opened 2026-08-27
docs: the plan for what shep owes its dogs

TurtIeSocks:
Ten tasks for what shep owes its dogs. Plan only, no code.

- `rehome` keeps `[dog.<name>]` instead of deleting the operator's own settings
- a lifecycle hook, so a dog can put things back before it is forgotten
- smits, which is what blocks shep-deploy's last task
- the reload deadline reaching the dog instead of being copied into it
- reserving the exit codes a dog returns through `shep <dogname>`
- a ruling on shep-client having no reconnect

Two of my own specifications were wrong and the plan says so.

The reload deadline cannot ride the reload response. `Response::Reloading` is a tuple variant under `#[serde(tag, content)]`, so a sibling field retypes `data` from array to object, and the handshake compares versions for equality. It goes on `ProcessInfo` instead, which is also the better home: the deadline is armed per replacement instance, so it closes all three inputs the dog infers rather than only the formula.

`exit.rs` does not owe new rows either. shep has no code path meaning "rolled back", so those variants would be dead surface. The collision is real though, since `shep <dogname>` passes a dog's exit code through unchanged. Section 9 reserves 12 and up with a test pinning the ladder.

Three bugs found while reading, none caused by this work, all fixed by tasks here:

- `shep-v1.md:451` gives `runtime` exit code 2, contradicting the table twelve lines up and the two paragraphs below it that explain why it is not 2
- two comments state `shep flock`'s column drop order backwards. The code, the width-46 test, the published 60-column render and the lookout's tiers all agree with each other and disagree with the prose
- three `web/` pages show `shep flock` without the EXIT column that shipped 2026-08-19

Only two dependency edges, so eight of the ten tasks parallelise.

### DIFF
--- a/docs/writing-plans/plans/2026-08-27-dog-prerequisites.md
+++ b/docs/writing-plans/plans/2026-08-27-dog-prerequisites.md
@@ -1,0 +1,2594 @@
+# Dog Prerequisites Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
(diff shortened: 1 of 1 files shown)


TurtIeSocks:
Closing this. A plan has no behaviour to review, so it does not want a PR of its own. The branch stays and the plan is its first commit; it will come back with the work it describes.

PR #45, opened 2026-08-28
Canary: docs-only change against the scoped CI

TurtIeSocks:
Proof run for #44. Only docs/releasing.md changes, so every Rust job should skip. Close without merging.


TurtIeSocks:
Cut from the wrong base: it carries #44's workflow edit, so the filter correctly reports rust=true and skips nothing. A real canary comes from main after #44 lands.

PR #204, opened 2026-09-11
feat: flexible flockfile env deserializing

TurtIeSocks:
Previously the AppConfig.env in Flockfiles only supported strings for all values, ie `"true"` instead of `true`. Now the deserializer is more flexible and will accept raw numbers and booleans.

Addressing CodeRabbit's review:

- **Whole numbers beyond `i64::MAX` now load**, not refused. Held in `i128`, the smallest type that carries both the full signed and the full unsigned range losslessly. Regression test pins `u64::MAX` to its exact string.
- **`env` values are redacted from `Debug`**. Both `EnvValue` and `EnvTable` now hand-implement `Debug`: `EnvValue` prints only its kind (`<str>`, `<bool>`, `<int>`, `<real>`), `EnvTable` prints only the key names and count. A `{:?}` on a config in the middle of a parse no longer drops secret-bearing values into a log. Both impls are pinned by exact-string tests so a lazy `derive(Debug)` fails loudly. The visitor methods and the two `deserialize` entry points that CodeRabbit's doc-coverage check flagged now carry docstrings as well.

Verification: 465 tests pass in `shep-core`, `cargo clippy --all-targets` clean.

## Summary by CodeRabbit

- **New Features**
  - Flockfile environment variables now accept strings, booleans, and whole numbers, converting supported values to strings automatically.
  - Supported consistently across TOML, YAML, JSON, and JSON5 formats.
  - Large unsigned integers are preserved without wrapping.
  - Floating-point values, arrays, and objects remain unsupported.
  - Environment variable values are redacted from debug output.

- **Documentation**
  - Updated the Flockfile reference with supported environment value formats and YAML scalar normalization details.


TurtIeSocks:
Fixed in `f40e40eb`. TOML's INTEGER type is signed 64-bit, so a TOML document cannot actually produce a value beyond `i64::MAX`; only JSON can. The `i128` storage is unchanged and still correct (it is simply invisible from a TOML Flockfile), but the comment and test docstring no longer claim TOML reaches that path.

TurtIeSocks:
@coderabbit review

TurtIeSocks:
@coderabbit full review

TurtIeSocks:
@CodeRabbit full review

PR #209, opened 2026-09-12
refactor(cli): consolidate command response plumbing

TurtIeSocks:
Group A of the qwen audit: CLI command response plumbing. 17 claims checked against the code, 13 acted on.

- `request_and_render` had 4 definitions and 9 more verbs inlining it by hand. One now, in `commands/rpc.rs`, plus `request_payload` for the verbs that render their own output
- `admin.rs` polled for teardown and emitted the same `KillRow` twice
- One predicate, match rule, emit and `#[cfg(unix)]` block in `dog_migration.rs`, `import/pm2/env.rs`, `secret.rs`, `runtime.rs`
- `dev_home` takes `Option<&Path>` instead of being handed `PathBuf::new()` as a sentinel
- `parse_selector_spec` in `commands/selector.rs`: `SelectorSpec::from(&parse_selector(..)?)` was written out 7 times
- Test setup hoisted in `admin.rs`, `trigger.rs`, `import/pm2/mod.rs`, `startup/unit.rs`

Left duplicated on purpose, reasons in the commits: `import::dotenv`'s two failure arms append which store was already written, and `query::flock`, `query::describe_selector` and `muster::muster` have success arms no single `Render` value expresses.

Two audit claims were wrong:

- The proposed predicate for `dog_migration` flips a De Morgan. `is_some_and(|t| !t.is_empty())` answers false for a non-table `Item` where the original answers true
- `fn test_streams() -> (Vec<u8>, Vec<u8>, Streams<'static>)` cannot compile, since `Streams<'a>` borrows the buffers it would be returned beside

`cargo test -p shep-cli` runs zero tests and exits 0. `shep-cli` is the crates.io placeholder at `crates/shep-cli-redirect/`. This directory is package `shep`. The audit plan lists that shape for groups B, E and F too.

Not done: `runtime.rs`'s `DISCOVERY_NAMES` duplicates `shep_core`'s private `DISCOVERY_ORDER`. They agree today, but the fix edits a file group D owns.

Verified: 1665 tests pass, `clippy --all-targets --all-features -D warnings` clean, doctests and `fmt --check` clean, and `cargo check --target x86_64-pc-windows-gnu` clean since three fixes sit on a platform split.

## Summary by CodeRabbit

- **Bug Fixes**
  - Standardized handling and reporting of unexpected daemon responses and request failures across CLI commands.
  - Improved consistency when resolving development-home paths, including cases where no home directory is available.

- **Refactor**
  - Consolidated command request, response rendering, selector parsing, and error-reporting behavior for more consistent CLI output.
  - Simplified runtime and secret-output handling without changing expected command behavior.

- **Tests**
  - Expanded coverage for development-home resolution, shell escaping, shutdown flows, and command execution scenarios.

### DIFF
--- a/crates/shep-cli/src/commands/admin.rs
+++ b/crates/shep-cli/src/commands/admin.rs
@@ -104,21 +104,14 @@ async fn kill_socket_free_with_wait(
     }
     #[cfg(unix)]
     {
-        if wait_for_socket_to_disappear(&paths.socket, wait).await {
-            write_outcome(emit(
-                &mut *streams.out,
-                streams.fmt,
-                "kill",
-                KillRow {
-                    pid,
-                    socket_removed: true,
-                },
-                streams.style,
-            ))
-        } else {
-            let message = "the shepherd was signalled, but teardown is still in progress";
-            streams.fail(ExitCode::DeadlineExceeded, message)
-        }
+        report_teardown(
+            streams,
+            &paths.socket,
+            pid,
+            wait,
+            "the shepherd was signalled, but teardown is still in progress",
+        )
+        .await
     }
     // Unreachable: `signal_graceful_stop` already returned on this platform.
     #[cfg(windows)]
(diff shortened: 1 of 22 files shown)


TurtIeSocks:
@coderabbitai review

PR #223, opened 2026-09-12
chore(ci): probe the paths filter with a Rust-only change

TurtIeSocks:
Throwaway draft to confirm .github/workflows/web.yml does not fire on a change outside its paths filter. Closing as soon as the checks list is visible.


TurtIeSocks:
Probe done. The docs site job did not run on this pull request, which is the intended behavior for a change outside the paths filter.

PR #224, opened 2026-09-12
fix(daemon)!: anchor a relative out_file/err_file at the app's cwd

TurtIeSocks:
`out_file` and `err_file` have always been documented as following `cwd` when relative. They did not. `assemble` built both with `PathBuf::from(render(...))` and joined nothing, and the shepherd is the process that opens them, so a relative path resolved against whatever directory the shepherd was launched from.

Reproduced against a real flock with `cwd = <dir>/b` and `out_file = "rel-out.log"`, loaded from `<dir>/a`:

- the sheep reported running in `<dir>/b`, its log landed in `<dir>/a`
- restarting the same Flockfile from `<dir>/c` moved the logs to `<dir>/c`, config unchanged
- `shep bleats relog --no-follow` from `<dir>/c` printed nothing, since the CLI opens the recorded path against its own directory

Either the join or the docs had to move. Went with the join: `lifecycle.rs` already states the rule for the sibling field, "the daemon resolves against its own cwd, so what crosses must be absolute", and `script` is absolutised by the CLI for exactly that reason. The log paths were the pair that never got it. Doing it in `assemble` rather than the CLI covers `shep set` and the override store too.

An app with no `cwd` runs in the shepherd's own directory, so a relative path there already resolves where the sheep runs and is left alone. That keeps `assemble` pure, with no `current_dir()` read.

Breaking: an app that set a relative `out_file` and a `cwd` will find its next log under the `cwd`.

- tests: four in `assemble.rs` for anchored, absolute, no-cwd and defaulted paths, plus one in `real_runner.rs` that spawns a real child and reads the line back out of the file under the app's `cwd`. Reverting the join fails that one.
- docs: `logs.astro` and the Flockfile reference now say where a relative path lands, and that reference's `out_file`/`err_file` defaults gained the `$SHEP_HOME` prefix they were missing. The generated CLI reference and `flockfile.schema.json` both regenerate unchanged.
- follow-up, not in this PR: with no `cwd` at all the recorded path stays relative, and `shep bleats` prints nothing rather than naming the file it could not open.
- worth a look on #220: `log_path_advisory` stats the raw config value, so whichever of the two lands second should hand it the anchored path.

## Summary by CodeRabbit

- **New Features**
  - Relative application log paths now resolve under the app’s configured working directory.
  - Absolute log paths and applications without a working directory continue to behave as before.

- **Documentation**
  - Updated log configuration guidance to clarify path resolution.
  - Documented the default locations for standard output and error logs.

### DIFF
--- a/crates/shep-daemon/src/assemble.rs
+++ b/crates/shep-daemon/src/assemble.rs
@@ -17,7 +17,7 @@
 use core::convert::Infallible;
 use core::fmt;
 use std::collections::BTreeMap;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
 use shep_core::config::ResolvedApp;
 use shep_core::config::template::{self, RenderError};
@@ -182,7 +182,8 @@ impl core::error::Error for AssembleError {
 /// Explicit `out_file`/`err_file` win over the default log path and render
 /// `{{instance}}` and `{{name}}` the way `env` and `args` do; normalize
 /// refuses a `{{secret:...}}` in either, and a path that collides across
-/// instances unless `merge_logs` asked for it.
+/// instances unless `merge_logs` asked for it. A relative one is anchored at
+/// the app's `cwd`.
 /// `SpawnSpec::stdin` carries `config.stdin` straight through: unlike
 /// `channel`, nothing else turns it on.
 ///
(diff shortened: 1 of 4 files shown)


TurtIeSocks:
@coderabbitai review

PR #228, opened 2026-09-13
fix(lookout): match a settings reply by its own ticket

TurtIeSocks:
The close dialog's fix in https://github.com/shep-pm/shep/pull/226 flagged the same assumption on the settings screen and left it alone. It is reachable, so this fixes it the same way.

- `Pending::Sent` eats no key, so `space` arms a second edit over an in-flight one and `Enter` sends it. `Escape` is the other door: it closes the screen without cancelling the write
- the first write's answer cleared the second's prompt and re-read `shep.toml`, which rebuilds the screen and throws the live edit away
- worse, a superseded refusal reopened the free-text editor over a live `Pending::Sent`, and the live reply then cleared that editor with `InputMode::Text` still on. Keyboard stranded until the operator finds `Escape`
- fixed by giving the write the `u64` ticket the config pane already mints and never reuses. `Settings::resolve` clears the prompt only for the reply that names it
- a reply it refuses still reports itself, it just touches no pending state. `Msg::DogWritten`'s `Ok` still asks the shepherd either way, or the file and the flock disagree

Four tests, all driving real keys through `App::update`. Mutated `Settings::resolve` back to ignoring the ticket and all four failed, restored.

Also pinned `InputMode::Text` on the reopened editor: deleting that line left the whole suite green.

No verb, key, flag, `shep.toml` key, exit code or JSON shape moves, so `web/` needs nothing. `lookout` is private, so none of the reshaped types are public API.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when multiple settings or dog edits are submitted in quick succession.
  * Prevented older write responses from clearing or overwriting the edit currently awaiting confirmation.
  * Preserved pending edits when responses arrive after the screen is closed or a write is refused.
  * Ensured settings are reloaded only when there are no active edits waiting for confirmation.

### DIFF
--- a/crates/shep-cli/src/lookout/app.rs
+++ b/crates/shep-cli/src/lookout/app.rs
@@ -298,6 +298,10 @@ pub enum Msg {
         /// The edit that was sent, echoed back: the cursor can have moved on
         /// while the write was in flight.
         edit: SettingEdit,
+        /// The ticket it went out with, echoed back: the screen can have
+        /// armed a second edit, or closed and reopened, while it was in
+        /// flight.
+        ticket: u64,
         /// Whether the write landed, or why it did not.
         result: Result<(), String>,
     },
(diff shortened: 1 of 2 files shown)


TurtIeSocks:
@coderabbitai review

TurtIeSocks:
@coderabbitai review

PR #230, opened 2026-09-13
perf(cli): compile the bleats match axis once, and share its chip suffix

TurtIeSocks:
Two changes to the lookout's match filter, folded together because they conflicted in all three files they share. #231 is merged in here rather than landing separately, which saves a CodeRabbit read and the rebase it owed this branch.

## The recompile

`Filters::match_ranges` parsed the match axis on every call, and `highlighted` calls it once per rendered line, so a bleats pane with a regex axis ran `Regex::new` on the same pattern once per line per frame.

Measured before changing anything, release profile, 80-line feed at 160x50, 48 body rows drawn:

| match axis | before | after | one `Regex::new` |
|---|---|---|---|
| literal `pool` | 21.9 us | 21.1 us | |
| `/pool\|ERROR/` | 599 us | ~25 us | 11.5 us |
| `/conn=[0-9]+/` | 1.0 ms | 26.4 us | 20.6 us |
| `/\d+\.\d+ms/` | 2.7 ms | 23.2 us | 43.6 us |
| `/\w+@\w+\.\w+/` | 27.0 ms | 24.8 us | 505.8 us |

`MIN_REDRAW` is 33 ms, so that last row spent 82% of a frame budget compiling one pattern 48 times, and typing it into the match box redraws once per keystroke. Compile cost scales with the pattern and the pattern is whatever an operator types, so the bound was theirs rather than ours.

- `Filters::matcher` is now `Option<Matcher>`, private, holding the typed text beside what it compiled to
- one field and one constructor, so a stale regex has nowhere to survive: not a replaced axis, not an abandoned edit, not a dropped chip
- `Matcher::new` returns `Option<Self>` and refuses empty text itself, because `is_match` and `ranges` disagreed about an empty needle and the guard was being held in three places
- `Matcher` hand-writes `Debug` (text and kind, since `Regex`'s own `Debug` reprints the pattern) and `PartialEq`/`Eq` (text alone), which is what the old comment gave as the reason not to cache
- callers read the axis through `Filters::match_text`
- an invalid regex still matches nothing and still never panics

Two of my first measurements looked cheap for the wrong reason: a matcher that survives no line is never highlighted, so `/(?i)error/` over a feed of INFO lines reported 83.8 us with zero rows drawn. Every number above is over a feed the pattern actually matches.

## The shared chip suffix, from #231

The three-arm suffix labelling a match chip as a regex or an invalid regex was written out once per pane, so the wording had to change in both places to stay in step. It is now `MatchKind::chip_suffix`, a `const fn` both call sites read. Only the suffix is shared: the embedded feed in `view::sheep` words its stream and level chips shorter on purpose.

Three conflicts on the merge, all the same shape, since #231 moved both panes to the helper in the same lines this branch made `Filters::matcher` private. Resolved by keeping their helper and reading it through `match_text`.

Ran #231's own mutation check against the resolution rather than only its tests, since this branch changes 58 lines in `bleats_full.rs` and their assertion compares a chip's own span. Setting `chip_suffix`'s `Literal` arm to `" (literal)"` fails all three of their tests, both pane ones included.

## Tests

`Matcher::new` increments a `#[cfg(test)]` thread-local counter. Two tests read it: setting the axis compiles once and reading it back over 80 lines compiles nothing, and a full redraw compiles nothing. Both confirmed to fail with the per-call compile put back, at 81 against 1 and 49 against 1, exactly the per-line counts. The redraw test also pins 48 body rows drawn and every one span-split, since the count alone passes over a pane drawing nothing.

The match-edit state machine had no coverage at all. It has ten now, covering every path that writes the axis, plus the `n` key's guard, whitespace-only text, and both hand-written impls. Each asserts the restored matcher still matches, not just that its text came back.

## Verification

Ten qwen rounds. Q raised 12 findings: 7 real and fixed, 5 dismissed, four of those on measured magnitude and one on a `thread_local` the chunked round could not see. My own rounds found 5 Q missed. Every counter and guard test was re-checked red against its own reversion.

Gate green: `cargo fmt --all --check`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `cargo test --workspace --all-features` (3875 passed), `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --all-features`, all exit 0.

No operator-visible change, so `web/` needs nothing. `lookout` is a private module in this crate's lib.rs, so the field going private is not a public API change.

Closes #231

## Summary by CodeRabbit

- **Improvements**
  - Match filters now process typed search text more efficiently, reducing repeated work during filtering and screen redraws.
  - Filtering behavior remains consistent while editing, applying, abandoning, and navigating match filters.
  - Match indicators now display consistent, clearer labels across the bleats and sheep views.
  - Empty match text is handled cleanly without producing an unnecessary literal match condition.

### DIFF
--- a/crates/shep-cli/src/lookout/app.rs
+++ b/crates/shep-cli/src/lookout/app.rs
@@ -4039,7 +4039,7 @@ impl App {
             KeyPress::MatchPrev => {
                 let stepping = self
                     .sheep_pane()
-                    .is_some_and(|pane| pane.feed().filters().matcher.is_some());
+                    .is_some_and(|pane| pane.feed().filters().match_text().is_some());
                 if stepping && let Some(feed) = self.sheep_feed_mut() {
                     feed.scroll_up(1);
                 }
@@ -4210,7 +4210,7 @@ impl App {
             KeyPress::MatchPrev => {
                 let stepping = self
                     .bleats_pane()
-                    .is_some_and(|pane| pane.filters().matcher.is_some());
+                    .is_some_and(|pane| pane.filters().match_text().is_some());
                 if stepping {
                     self.scroll_bleats_back(1);
                 }
@@ -4272,7 +4272,7 @@ impl App {
             KeyPress::Quit => Effect::Quit,
             KeyPress::TextChar(typed) => {
                 if let Some(pane) = self.bleats_pane_mut() {
-                    let mut text = pane.filters().matcher.clone().unwrap_or_default();
+                    let mut text = pane.filters().match_text().unwrap_or_default().to_string();
                     text.push(typed);
                     pane.set_match(text);
                 }
@@ -4280,7 +4280,7 @@ impl App {
             }
             KeyPress::TextBackspace => {
                 if let Some(pane) = self.bleats_pane_mut() {
-                    let mut text = pane.filters().matcher.clone().unwrap_or_default();
+                    let mut text = pane.filters().match_text().unwrap_or_default().to_string();
                     text.pop();
                     pane.set_match(text);
                 }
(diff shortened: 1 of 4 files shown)


TurtIeSocks:
@coderabbitai review

PR #235, opened 2026-09-13
docs(web): publish every secret subcommand's own flags

TurtIeSocks:
`shep secret --help` lists `set`, `get`, `unset` and `list` without their flags, so none of those flags reached the published reference. `--stdin` is the one that matters: it exists so a credential stays out of `ps` and shell history.

- Four quoted `VERBS` entries for the secret subcommands, the same treatment `import pm2` and `import env` got
- Regenerated the reference, 44 blocks to 48, and added the four to the Secrets group in `cli.astro`
- `every_visible_verb_reaches_the_docs_site_generator` walks the whole command tree now, not just the top level
- New `every_listed_verb_has_a_block_in_the_committed_reference` catches a stale generated file
- Dropped `VERB_NAMES`, 44 hand-kept lines. `cliReference.ts` reads its verb list off the `@@VERB:` markers instead
- Dropped `parseAliases`, dead code `astro check` flagged on every run
- Replaced the page's group count check with a by-name one, after the review found the count passes when a name is filed twice and another goes missing

`secret` and `import` are the only visible subcommand hosts. `dogs` takes flags rather than subcommands, so its own help already had them. `dog` and `daemon` are hidden and stay undocumented.

Every build-time check here is mutation checked, because a guard that passes either way is decoration:

| mutation | fails with |
| --- | --- |
| drop `"secret set"` from `VERBS` | `these verbs would be missing: ["secret set"]` |
| delete the `secret list` block from the reference | `the committed CLI reference has no block for ["secret list"]` |
| file `"secret set"` twice, drop `"secret get"` | `Missing: ["secret get"] ... Filed twice: ["secret set"]` |
| duplicate the `secret list` block | `Carried twice by the reference itself: ["secret list"]` |

The last two both passed before this PR.

Green on fmt, clippy, `cargo test --workspace --all-features`, rustdoc, `astro check` (0 errors, 0 warnings, 0 hints) and `astro build`.

## Summary by CodeRabbit

- **Documentation**
  - Expanded the CLI reference to include individual `secret` and `import` subcommands.
  - Improved CLI help organization and coverage for nested commands.
  - Updated the Secrets documentation section to list all available subcommands.

- **Bug Fixes**
  - Strengthened validation to detect missing, unknown, duplicate, or incomplete CLI reference entries.
  - Improved consistency between the available CLI commands and the committed documentation.

### DIFF
--- a/crates/shep-cli/src/cli.rs
+++ b/crates/shep-cli/src/cli.rs
@@ -1761,16 +1761,66 @@ mod tests {
         Cli::command().debug_assert(); // clap's own structural self-check
     }
 
-    /// fails when a visible verb is missing from the docs site's CLI
-    /// reference generator.
+    /// The generator's `VERBS` array, one entry per element, with a quoted
+    /// multi-word entry like `"secret set"` kept whole.
+    fn listed_verb_paths(generator: &str) -> Vec<String> {
+        let (_, rest) = generator
+            .split_once("VERBS=(")
+            .expect("the generator declares a VERBS array");
+        let (block, _) = rest.split_once(')').expect("the VERBS array closes");
+
+        // Splitting on the quote character puts every quoted entry at an odd
+        // index and everything unquoted at an even one.
+        block
+            .split('"')
+            .enumerate()
+            .flat_map(|(i, chunk)| {
+                if i % 2 == 1 {
+                    vec![chunk.to_string()]
+                } else {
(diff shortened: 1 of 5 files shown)


TurtIeSocks:
@coderabbitai review

TurtIeSocks:
@coderabbitai review

PR #244, opened 2026-09-13
fix(cli): refuse a home directory that is not valid UTF-8

TurtIeSocks:
`--home` and `$SHEP_HOME` arrive as an `OsString` and reach `ShepPaths::resolve` as a `String`, converted with `to_string_lossy`. A byte that is not valid UTF-8 becomes U+FFFD, and since U+FFFD is a legal filename character, shep then creates and uses a directory that is not the one you named. Nothing errors.

Refused at the door now, beside the relative-home gate from #232.

- `HomeRefusal::NotUtf8`, checked on `--home`/`$SHEP_HOME` and on the home directory when that is what supplies the root
- the relative gate still answers first when a path is both rootless and not UTF-8, so one bad home reports one reason
- `resolve`'s doc says why it takes a `String`, which is the question that started this
- tests are `cfg(unix)`; a Windows path is UTF-16 and its equivalent hole is an unpaired surrogate, not covered here

The reviewer who found this wanted `resolve` widened to an `OsString` lookup, matching `user_home` beside it. That does not work and it is not cheap.

It does not work because a home path does not stay a path. It reaches the `{{SHEP_HOME}}` template through `to_string_lossy`, `pipe_name` through `to_string_lossy`, and every log path on the wire as `Option<String>`. Widening the signature moves the loss downstream.

It is not cheap because `resolve` is public in a published crate with about twenty call sites, so it is a breaking change bought for a path shape macOS cannot even store: APFS refuses the filename with EILSEQ, measured. Linux allows it, which is why refusing beats ignoring.

### Two deliberate choices, so nobody tidies them away

The UTF-8 rule is not in `--home`'s clap doc, though #232 put its absolute-path rule there. `--home` is global, so that line repeats on all 48 verbs, and nobody types a non-UTF-8 path the way they type a relative one. `web/`'s cli page carries it instead. The two refusals are documented in different places on purpose.

The regenerated CLI reference in this branch is not from this change. #232's absolute-path sentence landed in the doc comment without the reference being regenerated, so main has been serving a copy that does not match its own binary. Four lines, picked up on the way past.

### Also here

Two tests for `pipe_name`'s 64-byte truncation and the trim that has to follow it. Both existing tests use a home far under the cap, so the slice and the second `trim_end_matches` had never run. Checked by removing the trim: the stem keeps its dash and doubles it against the digest's own.

Every guard here was verified by removing it and watching a test fail.

## Summary by CodeRabbit

- **Bug Fixes**
  - The CLI now clearly rejects `--home` or `SHEP_HOME` paths containing invalid UTF-8, with an actionable usage error instead of silently replacing characters.
  - Pipe names remain distinct for home directories that share the same shortened name.

- **Documentation**
  - Updated the CLI reference to document the UTF-8 requirement for home paths and explain where the path is used.


TurtIeSocks:
@coderabbitai review

PR #259, opened 2026-09-15
test: standardise the example service name in fixtures and docs

TurtIeSocks:
Standardises the example service name across the repo to `api-auth`.

- `selector.rs` and `output/table.rs` fixtures
- `supervisor.rs` uses `zone`, not `api`: it has to sort after `worker` or name order collapses onto settle order. An assertion pins that now
- two design docs drop the name instead of exchanging it, since those passages record real incidents and an example name would have attached a measured 200 KB to a file that never existed

Swept all 65 occurrences for the same sort-order hazard. Only `supervisor.rs` had it.

One change is not a rename. That test's doc comment said the rows come out as `[1, 0, 2]`, which is settle order. They come out in name order, `[0, 2, 1]`, which the assertion below it has always demanded. `spawn_trigger_task` sorts by `(name, id)` and says so in its own comment. The doc comment now matches.

Partial. `cli.rs`, `lifecycle.rs`, `snapshot.rs`, `map.md` and the regenerated CLI reference land in a second PR once #258 and the shep-daemon split merge.


TurtIeSocks:
Closing. These four are landing inside the shep-daemon split instead.

PR #304, opened 2026-09-16
refactor(paths): centralize home directory variable

### LINKED ISSUE #290: `HOME_DIR_VAR` is defined twice
`HOME_DIR_VAR` holds the name of the home directory environment variable, `$HOME` on unix and `%USERPROFILE%` on Windows. Two crates declare it, byte-identical, `pub(crate)` in each, behind the same pair of `cfg` arms. Either copy can change without the other.

## Every site

```
crates/shep-core/src/paths.rs:85    #[cfg(not(windows))]
crates/shep-core/src/paths.rs:91    #[cfg(windows)]
crates/shep-cli/src/home.rs:453     #[cfg(not(windows))]
crates/shep-cli/src/home.rs:462     #[cfg(windows)]
```

## How the set was checked

```bash
rg -n 'HOME_DIR_VAR' crates/
```

Four declarations and nothing else. This issue originally named `crates/shep-cli/src/lib.rs`, which has no occurrence of the constant at all.

## Precise details

```
fix: make one the source of truth. shep-core's is the better home, since shep-cli already depends on it.
value: one declaration rather than two, and the cfg pair stops being written twice.
breaking: none. Both are pub(crate).
risks: none.
already shared: nothing. These are two independent declarations of the same constant.
```

Voyagerroc-Lab:
## Summary
- make the platform-specific HOME_DIR_VAR constant a single source of truth in shep-core
- reuse it from shep-cli home and dev commands
- remove the duplicate platform definitions

Closes #290

## Validation
- git diff --check
- Cargo tests were not run because Cargo is unavailable in the execution environment.

## Summary by CodeRabbit

* **Refactor**
  * Consolidated platform-specific home-directory environment variable handling into shared tooling.
  * Preserved existing `shep dev` behavior, fallback paths, and diagnostic messages across supported platforms.
  * Home-directory resolution continues to use `$HOME` on Unix-like systems and `%USERPROFILE%` on Windows.


TurtIeSocks:
@Voyagerroc-Lab please address the findings from CodeRabbit's review and then this should be good to go. Thanks for the contribution!

TurtIeSocks:
@coderabbit full review

TurtIeSocks:
Thanks for this, and for coming back to fix the duplicate yourself.

Pushed one commit on top restoring the doc comment on the Windows spelling. The rewrite had left "a stock Windows session sets none of the first", which dropped the `user_home` link and the reason `%USERPROFILE%` gets named rather than `HOME`. Nothing would have caught that locally, since the item is `#[cfg(windows)]` and only the windows-gnu cross-check compiles it.

Merging now.

PR #305, opened 2026-09-16
fix(cli): match COLORTERM and TERM case-insensitively

### LINKED ISSUE #282: `deep_colour_terminal` misses `COLORTERM=TrueColor`
`crates/shep-cli/src/style.rs:98`:

```rust
value.contains("truecolor") || value.contains("24bit")
```

`COLORTERM=TrueColor` therefore reads as no deep colour. The test below it only ever passes lowercase, so nothing fails.

Fix is a case fold on the variable's value. Worth checking `TERM`'s `256color` match in the same function while you are there.

maskjelly:
Closes #282.

`deep_colour_terminal` compared exactly: `$COLORTERM` had to contain a lowercase `truecolor` or `24bit`, and `$TERM` a lowercase `256color`. A terminal spelling it `COLORTERM=TrueColor` therefore read as no deep colour and got the 16-colour fallback — and the test beside the function only ever passed the lowercase spellings, so nothing failed.

Both values are folded before matching now, and the test pins `TrueColor`, `24BIT` and an upper-case `$TERM`.

**Fail-without-fix.** With the fold removed and the extended test in place:

```
assertion failed: deep_colour_terminal(None, Some(OsStr::new("TrueColor")))
test result: FAILED. 1 passed; 1 failed; 0 ignored
```

**Gate** (all green): `cargo +1.93 fmt --all -- --check`, `cargo +1.93 clippy --workspace --all-targets --all-features -- -D warnings`, `cargo test --workspace --all-features`, and `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --all-features --document-private-items`.

Two notes on the toolchain: I used the CI's pinned 1.93 for fmt and clippy, because the active stable here is 1.98 and its clippy reports 13 lints on pristine `main` (ten `unused_imports` in `lookout/app/*` test modules and one `chunks_exact_to_as_chunks` in `serve/auth.rs`) that have nothing to do with this change; 1.93 is clean on both trees. And I left `web/` alone deliberately — the docs trigger lists verbs, flags, config keys, exit codes and payload shapes, and terminal colour detection is none of those; please say so if you read it otherwise.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved terminal color detection to recognize mixed- and uppercase color capability values.
  * Ensures truecolor, 24-bit, and 256-color terminals are detected consistently regardless of capitalization.


TurtIeSocks (approved):
Thank you for the contribution!

PR #532, opened 2026-09-19
refactor(cli): output/rows/dogs.rs: three identical row builders

### LINKED ISSUE #384: shep-cli: `output/rows/dogs.rs`: four dog-action row types are hand-copied
`crates/shep-cli/src/output/rows/dogs.rs` carries five idents with the same four fields, and four `Render` impls that are functionally identical.

| ident | name | source | shepherd_acted | status |
|---|---|---|---|---|
| `DogEnabledRow` | `String` | `DogSource` | `bool` | `String` |
| `DogDisabledRow` | `String` | `DogSource` | `bool` | `String` |
| `DogAdoptedRow` | `String` | `DogSource` | `bool` | `String` |
| `DogRehomedRow` | `String` | `Option<DogSource>` | `bool` | `String` |
| `DogActionRow<'a>` | `&str` | `&str` | `bool` | `&str` |

What actually varies across the four `Render` impls, roughly 200 lines:

- `headers`, `rows_for`, `JSON_ONLY` and `PRIORITIES`: nothing, all four delegate to `DogActionRow`
- `rows()`: nothing, except `rehome` resolving its `Option` to `-`
- `json_key_for`: a string literal that reaches only a panic message
- the doc comments: these genuinely differ per verb and are worth keeping

`dog_action_json_key!($caller, $header)` uses `$caller` only in `panic!("{caller}::headers() does not include {other:?}")`. The key mapping is the same in all four, and all four `headers()` return `DogActionRow::headers()`, so that arm is unreachable the same way each time.

That makes the hand-written type name a latent bug as well as noise. A copied impl that keeps the previous string panics under the wrong type name and nothing checks it. All four are correct today by luck, not by construction.

Worth preserving:

- four distinct public types. The JSON output distinguishes them and the drift tests at `dogs.rs:461` construct them by name
- the per-verb doc comments
- `rehome`'s `Option<DogSource>`, which is a real difference

A `dog_action_row!` macro taking the docs, the type name and the source field type could generate both the struct and its impl, with `stringify!` replacing the hand-written name. Four invocations for the lot. Other shapes may fit better, that one is just what the variation suggests.

#532 tried a shared builder taking `(name, source, acted, status)` and was closed. It attacked the identical `rows()` bodies, which are a symptom: the four values still got written at each site, positionally instead of by name, so it traded a compile-time guarantee for three lines. The duplication is four hand-copied types, and no builder function reaches that.

TurtIeSocks:
Closes #384.

Written by the local model from the issue text, gated here with cargo fmt, clippy and shep's tests. 1 commit, 59 lines, 1 file.

Labelled for review: it changes 27 of the 28 lines the issues cite; the first attempt failed its gate.

## Summary by CodeRabbit

* **Refactor**
  * Streamlined internal row construction for dog action output.
  * Preserved existing output behavior and formatting across enabled, disabled, and adopted dog actions.


TurtIeSocks:
Closing this. The finding is real, but the fix costs more than the duplication does.

The three bodies were identical because their four field values are, and those are still written out at each of the three sites after this change. What actually goes away is three lines and the named fields. `build(name, source, shepherd_acted, status)` takes three `&str` and a `bool` positionally, so swapping `name` and `source` compiles and renders wrong, where the struct literal made that impossible. It also leaves two ways to build one row, `rows()` and `build()`.

Nothing wrong with the execution, it is what #384 asked for. Closing that too.

PR #536, opened 2026-09-19
refactor(daemon): two of the three duplicated patterns in the snapshot tests

### LINKED ISSUE #504: Route every FlockSnapshot literal through one constructor, and share the list-sort-project sequence
Three duplicated patterns around the daemon's snapshot tests, beyond the async scaffold already filed elsewhere. Two are being fixed in https://github.com/shep-pm/shep/pull/536. The third was measured and declined; see the bottom of this issue.

`FlockSnapshot { version: SNAPSHOT_VERSION, saved_at_ms: 0, apps: ... }` is spelled out inline 18 times across 10 files. `graph.rs` already has `roll_of()` doing the same thing for the common case of every app running one instance, and one of the inline sites sits in that same file below `roll_of`'s own definition. Several sites outside shep-daemon hardcode `version: 1` instead of naming `SNAPSHOT_VERSION`, because that constant is `pub(crate)`.

The sequence "list, sort by name, project to `Vec<(&str, ProcStatus)>`" is written out verbatim four times.

## Every site

Line numbers are as of `main` before https://github.com/shep-pm/shep/pull/536.

```
crates/shep-daemon/src/snapshot/tests/muster.rs             6 FlockSnapshot literals
crates/shep-daemon/src/boot/tests/dogs.rs                   3
crates/shep-cli/src/commands/query/secret_inspection.rs     2   hardcodes version: 1
crates/shep-daemon/src/snapshot/mod.rs                      1   in the parent tests module
crates/shep-daemon/src/snapshot/tests/graph.rs              1   below roll_of's own definition
crates/shep-daemon/src/snapshot/tests/secrets.rs            1
crates/shep-daemon/src/boot/tests/readiness.rs              1
crates/shep-daemon/src/boot/running.rs                      1   production, not a test
crates/shep-cli/src/secret_readers.rs                       1   hardcodes version: 1
crates/shep-cli/src/commands/query/testing.rs               1   hardcodes version: 1

crates/shep-daemon/src/snapshot/tests/graph.rs:111-116      list, sort, project
crates/shep-daemon/src/snapshot/tests/muster.rs:147-150     list, sort, project
crates/shep-daemon/src/snapshot/tests/muster.rs:208-211     list, sort, project
crates/shep-daemon/src/snapshot/tests/muster.rs:276-279     list, sort, project
```

## How the set was checked

```bash
git grep -n 'FlockSnapshot\s*{' -- 'crates/**/*.rs'
rg -n '\.list\(\)\.await' crates/shep-daemon/src/snapshot/tests/
```

The first command returns 23 raw matches. Three are the struct definition and two return types, and two more are the canonical builders themselves, `FlockRegistry::roll` and `roll_of`, which are what the other 18 should be calling. So 18 duplicated literals. This issue previously said eight, because the grep behind that number was scoped to `crates/shep-daemon/src/snapshot/tests/` rather than the tree, and then said 23, which counted the matches rather than the duplicates.

One correction to the original's fix suggestion, still standing: there is no `TestPaths` type in this codebase. The parameter is `ShepPaths`.

## The third finding, measured and declined

The original asked for `spawn_supervisor`'s `paths: ShepPaths` to become `&ShepPaths` so its callers could drop a `.clone()`. Do not do this.

`spawn_supervisor` hands `paths` straight to `SupervisorBuilder::new`, and the actor owns it from there, so a reference does not remove the clone. It moves the clone inside the function body, where every caller pays it.

```
call sites                 92
  pass paths.clone()       14   would lose a clone
  move an owned value      78   would gain one
```

The 78 mostly pass `test_paths(&dir)`, a temporary that moves straight in today. So the change is 64 more clones of a struct of paths, 92 call sites edited, and a major version bump on a published crate, in exchange for the ergonomics. Filing it fresh is fine if someone thinks that trade is worth making, but it should be filed as that trade rather than as a cleanup.

## Precise details

```
fix: FlockSnapshot::with_apps(Vec<SavedApp>), used at all 18 sites, and point the graph.rs site at roll_of() like every other call in that file; a fn sorted_status(&handle) -> Vec<(String, ProcStatus)> helper for the list/sort/project sequence.
value: two fixes across ten files. The with_apps constructor also stops three sites outside shep-daemon hardcoding version: 1 because SNAPSHOT_VERSION is pub(crate).
breaking: none. Both remaining fixes are additive or test-only. The spawn_supervisor signature change would have been breaking, which is part of why it was measured before being attempted, and it is not being made.
risks: none.
already shared: graph.rs already has roll_of() for the common one-instance case; it just was not used at every site.
checked with: cargo test --workspace --all-features
```

TurtIeSocks:
- `FlockSnapshot::with_apps` builds a roll at the current `SNAPSHOT_VERSION`, and `crate::testing::roll_of` builds one where every app is recorded as running
- 18 sites stopped spelling out all three fields, 14 in shep-daemon and 4 in shep-cli
- `crate::testing::sorted_status` replaces the list, sort, project sequence in four flock assertions
- `spawn_supervisor`'s signature is unchanged, see below

The issue says 8 `FlockSnapshot` literals in the snapshot tests. It is 9: the grep ran over `snapshot/tests/` and missed the one still inline in the parent tests module in `snapshot/mod.rs`. Four more in the boot tests and four in shep-cli are the same shape, so they went too. The shep-cli ones were writing `version: 1` by hand, because `SNAPSHOT_VERSION` is `pub(crate)` and a caller in another crate cannot name it.

Skipped the third pattern. `spawn_supervisor` stores `paths` in the actor, so taking `&ShepPaths` just moves the clone inside the function. 14 of its 92 call sites clone today and the other 78 move an owned `test_paths(&dir)` straight in, so the change removes 14 clones and adds 78, and costs a major bump on a published crate. Say the word if you want it anyway.

Gate green locally on macOS: fmt, clippy, `cargo test --workspace --all-features`, and rustdoc with `--document-private-items`.

Resolves #504

## Summary by CodeRabbit

- **Refactor**
  - Standardized snapshot creation through shared construction helpers.
  - Centralized common test-fixture setup and status reporting.
  - Snapshot records now consistently use current version metadata and preserve saved application details.

- **Tests**
  - Updated daemon, CLI, secret, restoration, readiness, and snapshot tests to use shared fixtures.
  - Existing test scenarios, assertions, application data, and process-status outcomes remain unchanged.


TurtIeSocks:
@coderabbitai full review

TurtIeSocks:
@coderabbitai full review

PR #538, opened 2026-09-20
fix(web): surface a visible failure when a code-block copy fails

### LINKED ISSUE #432: web: a failed clipboard write leaves CodeBlock's button saying nothing went wrong
- files: `web/src/components/docs/CodeBlock.astro`
- findings: `web/src/components/docs/CodeBlock.astro:202-204`, `catch {
  return; }` on `navigator.clipboard.writeText(text)`: the button stays in
  its default "copy" state with no indication the copy failed. Confirmed
  current.
- fix: on catch, flip the button into a visible error state (e.g. textContent
  "failed", `aria-label="Copy failed"`).
- value: this is a real UX defect, not duplication: a user who hits a
  clipboard permission denial or a non-secure context believes the copy
  worked when it silently did not. Highest-value item in the web slice.
- breaking: none, front-end only.
- risks: none.

ump45nose:
## Short summary

Closes #432.

On `web/src/components/docs/CodeBlock.astro` the copy button's click handler bailed out silently in two failure paths: a rejected `navigator.clipboard.writeText()` (permission denial) and a missing clipboard API (non-secure context). In both, the button stayed in its default "copy" state, so a reader had no way to know the command was not copied.

The handler now flips the button into a visible failure state in both paths — label `failed`, `aria-label="Copy failed"`, copy icon restored, copied icon hidden. Unlike the copied state this does not auto-revert after 1800 ms; it clears on the next attempt, so the failure stays visible until the user retries.

## Precise details

- Repo/branch: `ump45nose/shep`, branch `fix/codeblock-copy-error-state`, commit `a8ae063`, base `main` (upstream `01110f0`).
- Files touched: `web/src/components/docs/CodeBlock.astro` only (+17/-1).
- Behaviour table:
  - clipboard API missing → label `failed`, aria-label `Copy failed` (was: silent return, default state)
  - `writeText` rejects → label `failed`, aria-label `Copy failed` (was: silent return, default state)
  - `writeText` resolves → unchanged: label `copied`, auto-revert after `REVERT_DELAY_MS`
- Gate run locally, in order, per CONTRIBUTING: `cd web && npm ci` (clean install), `npx astro check` → `Result (74 files): 0 errors`, `npm run build` → passes.
- Not tested: real-browser interaction (no browser harness here); the failed-clipboard path was verified by reading the handler, not by denying a permission in a live session. CI's Linux/Windows builds do not exercise `web/`; the `astro check` half of the web gate is the part that type-checks this change.
- No ported source from pm2 or any other process manager (front-end only, none involved).
- Written with AI assistance (ZCode/GLM); every line above is explainable — the state flip mirrors the existing `setCopied` icon/label/aria handling minus the revert timer, because a failure is not transient feedback.

## Summary by CodeRabbit

- **Bug Fixes**
  - Code-block copy controls now clearly indicate when copying fails, including when clipboard access is unavailable or an error occurs.
  - Failure feedback remains visible until the next copy attempt.


TurtIeSocks:
@coderabbitai full review

web/src/components/docs/CodeBlock.astro
```diff
@@ -195,12 +195,28 @@ const headerCopy = Boolean(label) && iconCopy;
...
+      // the button claiming nothing went wrong. Unlike the copied state this
+      // does not auto-revert: the failure stays visible until the next try.
+      const setFailed = () => {
+        if (label) label.textContent = "failed";
+        if (copyIcon) copyIcon.style.display = "";
+        if (copiedIcon) copiedIcon.style.display = "none";
+        if (copyIcon || copiedIcon) {
+          button.setAttribute("aria-label", "Copy failed");
+        }
+      };
```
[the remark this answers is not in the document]
TurtIeSocks: Confirmed, this one is real. A successful copy arms `revertTimer` and `setFailed()` leaves it running, so a failure inside the next `REVERT_DELAY_MS` reverts itself to `copy` with no second attempt. That is the opposite of what the comment just above `setFailed` promises.

`CopyCloneButton.astro` on main already clears the timer in its `catch` if you want a reference. Same fix here and this is good to go.

ump45nose:
Fixed in f83883ea. The finding was valid and is exactly the race the comment above `setFailed` claims cannot happen: a successful copy inside the previous 1.8s window leaves its revert timer pending, and that timer's whole job is to wipe the button back to resting state — so a second click that now fails gets its `failed` state cleared before it can be read. `setFailed` clears the pending revert and resets the handle before applying the failure UI.

Web gate re-run on the fix: `npm ci`, `astro check` (0 errors, 0 warnings, 0 hints), `npm run build` (all 11 verifier tests pass).

TurtIeSocks:
@coderabbitai review

TurtIeSocks:
The timer clear is the right catch, and that comment earns its length. `setCopied(false)` firing over the top of `failed` is the kind of thing you only find by reading the interleaving.

One mode it does not reach though. `headerCopy` renders icons and no `data-copy-label` at `:44`, and the other branch renders the label and no icons at `:90`, so the two are exclusive. In icon mode `setFailed` sets `copyIcon` to `""` and `copiedIcon` to `"none"`, which is the resting state, and `label` is null. `aria-label` is the only thing that changes, and swapping that on an already-focused button does not reliably announce.

`community-dogs.astro:133` and `:138` are both `iconCopy`, so the two blocks with the icon button are the two where a failure shows nothing. Worth a third SVG, or a class on the button?

TurtIeSocks:
@coderabbitai review

web/src/components/docs/CodeBlock.astro
```diff
@@ -195,12 +195,35 @@ const headerCopy = Boolean(label) && iconCopy;
...
+        // its resting state: left alone it fires straight over the top of
+        // the failure, so "failed" flashes away unread. Clearing it here is
+        // what makes the no-auto-revert promise above true.
+        clearTimeout(revertTimer);
+        revertTimer = undefined;
+        if (label) label.textContent = "failed";
+        if (copyIcon) copyIcon.style.display = "";
+        if (copiedIcon) copiedIcon.style.display = "none";
+        if (copyIcon || copiedIcon) {
+          button.setAttribute("aria-label", "Copy failed");
```
TurtIeSocks, line 213: The text variant never hears about any of this. Line 96 gives that button a static `aria-label="Copy code"`, which overrides its own text as the accessible name, and the `if (copyIcon || copiedIcon)` guard is false there because that branch renders no icons. So a screen reader is told "Copy code" through copy, copied and failed alike, and the failure this PR exists to surface is the one state it cannot report.

Three changes, and the first is what makes the other two safe:

```js
// beside `revertTimer`, so the resting name is whichever variant rendered
const restingLabel = button.getAttribute("aria-label") ?? "Copy";

// setCopied, guard dropped
button.setAttribute("aria-label", copied ? "Copied" : restingLabel);

// setFailed, guard dropped
button.setAttribute("aria-label", "Copy failed");
```

Hardcoding `"Copy command"` in `setCopied` instead would rename the text variant's button on every revert, which is why the resting name gets hoisted rather than repeated.

Not a `suggestion` block because `setCopied`'s guard sits at line 194, just outside this PR's only hunk, and half the fix would leave `Copy failed` stuck there after the next success.

Built it at `f83883ea` and drove both paths in a browser:

```
resting   Copy code    / copy
failure   Copy failed  / failed
success   Copied       / copied
revert    Copy code    / copy
```

`npx astro check` clean, `npm run build` green.

TurtIeSocks:
@coderabbitai review

web/src/components/docs/CodeBlock.astro
TurtIeSocks: This one turns out to be pre-existing on main, tracked in #581 with the verified fix. Your setFailed inherits the same guard rather than introducing it, so nothing needed here.

PR #539, opened 2026-09-20
fix(web): single-source the hero install command and await its clipboard write

ump45nose:
## Short summary

Closes #433.

`web/src/components/landing/CopyCloneButton.astro` had the two problems the issue describes, fixed as two commits so either can be reverted alone:

1. **Content drift** — the install command existed twice: the `INSTALL_COMMAND` constant (feeding `data-clone-command`) and a hardcoded copy of the same string as the button's visible text. The visible text is now rendered from the same constant (`{INSTALL_COMMAND}`), so what the button shows and what it copies can no longer desynchronise.

2. **Silent false success** — the click handler fired `navigator.clipboard?.writeText(command)` fire-and-forget and immediately flipped the label to `copied`, regardless of whether the promise resolved. The handler is now `async`, awaits the write, and only shows `copied` on fulfilment. On rejection (denied permission, missing clipboard API / non-secure context) it shows `failed` until the next attempt, mirroring the docs CodeBlock failure state from #538. The pending revert timer is cleared on failure so it cannot stomp the failed label.

## Precise details

- Repo/branch: `ump45nose/shep`, branch `fix/copyclonebutton-clipboard-promise`, commits `792cd03` + `6d01720`, base `main` (upstream `01110f0`).
- Files touched: `web/src/components/landing/CopyCloneButton.astro` only (+15/-4 across both commits). Header doc comment updated to mention the failed state.
- Behaviour table:
  - visible text: rendered from `INSTALL_COMMAND` (single source; `data-clone-command` unchanged)
  - write fulfils → label `copied`, auto-revert after `REVERT_DELAY_MS` (unchanged)
  - write rejects or clipboard API missing → label `failed`, no auto-revert (was: label `copied` regardless)
- Gate run locally, per CONTRIBUTING, after the full change: `npx astro check` → `Result (74 files): 0 errors`; `npm run build` → all verifier checks `fail 0`, exit 0. (`npm ci` state from the #538 run in this clone.)
- Not tested: real-browser interaction; the rejection path was verified by reading the handler, not by denying clipboard permission live. CI's Linux/Windows builds do not exercise `web/`.
- No ported source from pm2 or any other process manager (front-end only, none involved).
- Written with AI assistance (ZCode/GLM); rationale for each line is stated above and in the commit bodies.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved install-command copying feedback.
  - The button now clearly indicates when copying succeeds or fails.
  - Failure status remains visible until the next copy attempt.

PR #540, opened 2026-09-20
fix(web): use constant for CopyCloneButton text and await clipboard write

### LINKED ISSUE #433: web: CopyCloneButton duplicates the install command and never awaits the clipboard
- files: `web/src/components/landing/CopyCloneButton.astro`
- findings:
  - `web/src/components/landing/CopyCloneButton.astro:7,11`, the install
    command `"cargo install shep"` is written twice: once as the
    `INSTALL_COMMAND` constant (fed to `data-clone-command`), once hardcoded
    as the button's visible text.
  - `web/src/components/landing/CopyCloneButton.astro:77` , 
    `navigator.clipboard?.writeText(command)` is fire-and-forget, no `await`
    or `.catch`; the label flips to "copied" regardless of whether the
    promise resolves.
- fix: render the visible text from the same `INSTALL_COMMAND` constant
  (split into a prompt span plus an interpolated command); `await` the
  clipboard write (or attach `.catch`) and only show "copied" on fulfilment.
- value: one file, two real problems (a content drift risk and the same
  silent-failure shape as the CodeBlock.astro finding above).
- breaking: none, front-end only.
- risks: none.

kragent66-glitch:
## Problem

Fixes #433

Two issues in `web/src/components/landing/CopyCloneButton.astro`:

1. **Duplicated install command**: Line 11 hardcodes `cargo install shep` instead of using the `INSTALL_COMMAND` constant defined at line 7. If the constant is updated, the visible text becomes stale.

2. **Fire-and-forget clipboard**: `navigator.clipboard?.writeText(command)` has no `await` or `.catch`, so the label flips to "copied" immediately regardless of whether the clipboard write succeeded.

## Approach

1. Render the visible text from the constant: `{INSTALL_COMMAND}` instead of hardcoded `cargo install shep`.

2. Make the click handler `async`, `await` the clipboard write inside a `try/catch`, and only show "copied" on success.

## Verification

- `npx astro check`: 0 errors, 0 warnings, 0 hints
- `npm run build`: all verify-* scripts pass, 0 test failures
- Button renders "$ cargo install shep" (unchanged visible output)
- Clipboard failure no longer shows false "copied" state

## Files Changed

- `web/src/components/landing/CopyCloneButton.astro` — 2 targeted changes (12 insertions, 8 deletions)

## Known Issues

None. The fix is fully verified locally.


TurtIeSocks:
Thanks for this, and sorry about the duplicate: #539 landed on the same issue about ten minutes ahead of yours and is merged now.

One difference worth keeping in mind for next time. `navigator.clipboard?.writeText(command)` short-circuits to `undefined` when the API is missing, so the `await` resolves, nothing throws, and the label still says `copied` in a non-secure context. Dropping the `?.` lets the TypeError land in your `catch` instead.

Plenty more open if you fancy another, there are a few under `good first issue`.

PR #541, opened 2026-09-20
refactor(cli): take json_key panic names from type_name

### LINKED ISSUE #533: shep-cli: output/rows: seven hand-written type names the compiler already knows
- files: `crates/shep-cli/src/output/rows/dogs.rs`, `crates/shep-cli/src/output/rows/replies.rs`
- findings: seven `json_key_for` impls pass their own type's name to a macro as a string literal, and nothing checks it against the enclosing type. `dogs.rs:253,302,353,407` via `dog_action_json_key!`, `replies.rs:119,212,266` via `reply_rows_json_key!`. All seven are correct today. The literal is only ever read by the `other =>` panic arm, which is unreachable while `headers()` and the match agree, so a rename would compile clean and leave the wrong name in a message nobody sees until it fires.
- fix: drop the `$caller` parameter from both macros and use `core::any::type_name::<Self>()` inside them. Both expand inside `impl Render for X`, so `Self` is in scope at all seven sites and `type_name` returns `&'static str` already.
- value: one fewer thing a rename can silently falsify, and seven call sites get shorter.
- breaking: none, the macros are private to their modules.
- risks: `type_name` renders a qualified path, so the panic message gains a module prefix. Worth checking no test pins the current string.

Not duplication as such. The two macros look alike but their header maps genuinely differ, so they should stay two. Related to https://github.com/shep-pm/shep/issues/384, which covers the row types in `dogs.rs` themselves.

Found while validating a batch of duplication issues, then confirmed by reading all seven sites.

ump45nose:
## Short summary

Closes #533.

Seven `json_key_for` impls in `dogs.rs` and `replies.rs` passed their own type's name into a private macro as a string literal. Nothing checked it against the enclosing type, so a rename would compile and leave the wrong name in the unreachable `other =>` panic.

Both macros now drop the `$caller` parameter and use `core::any::type_name::<Self>()` instead. They still expand inside `impl Render for X`, so `Self` is in scope at all seven sites.

## Precise details

- Repo/branch: `ump45nose/shep`, branch `refactor/json-key-macros-type-name`, commit `4f9ec9e`, base `main` (upstream `01110f0`).
- Files touched: `crates/shep-cli/src/output/rows/dogs.rs`, `crates/shep-cli/src/output/rows/replies.rs` (+19/-17).
- Behaviour: JSON keys are unchanged. The panic message now carries a qualified path from `type_name` rather than the bare type name. No test pinned the old string.
- Tests: `cargo fmt --all --check` pass. `cargo test -p shep --lib --all-features output::rows` → 91 passed. `cargo test -p shep --lib --bins --all-features -- --skip ::slow::` → 2178 passed; 7 failed in `commands::hook::tests` with `TimedOut { after: 2s }` (process-spawn timeouts, unrelated to these macros). `cargo clippy -p shep --all-targets --all-features -- -D warnings` fails on pre-existing unused imports in `lookout/app/*` and a `chunks_exact` lint in `serve/auth.rs`; none of those files are in this diff.
- Not tested: full workspace `cargo test --workspace --all-features` and `cargo doc` (CONTRIBUTING gate); Linux/Windows CI. Local linking needed `SDKROOT` pointed at MacOSX15.4.sdk because the default MacOSX27.0.sdk TBD files fail to link on this machine.
- No ported source from pm2 or any other process manager.
- Written with AI assistance (Grok); every line is explainable — the macros still exist because rustc's dead-code pass cannot see a shared fn used only inside other trait impls, and `type_name::<Self>()` is the compiler's name for the type the literals were trying to keep in sync.

## Summary by CodeRabbit

* **Refactor**
  * Simplified internal JSON key generation for dog-action and reply rows.
  * Error messages now automatically reflect the relevant type name, improving accuracy when types are renamed.
* **Behavior**
  * JSON key mappings and panic behavior remain unchanged.


TurtIeSocks:
@coderabbitai full review

PR #543, opened 2026-09-21
ci(release): report the breaking-bump verdict on the release PR

TurtIeSocks:
## Short summary

The guard from #542 runs on a push to main, not on the release pull request, so a refusal was a red run in the Actions tab and nothing on the pull request itself.

- writes the verdict as a `release bump` commit status on the release pull request's head
- written on both outcomes, and before the exit that fails the run
- needs `statuses: write`, which this job did not hold

## Precise details

Measured 2026-09-21. The guard failed on `aee84c99` while #531 read `MERGEABLE / CLEAN` with 31 green checks, none of which was the guard. Nothing on the page said the version was wrong. It informed, it did not block.

A status attaches only to the sha it names, so release-plz's next force-push leaves the new head carrying none. That reads as unverified rather than as inherited approval, which is the right default here: the workflow that force-pushes is the same one that writes the status, so a head without one is a head the guard has not run against yet.

Adding `release bump` to main's ruleset is what turns this from a report into a block. Left out on purpose. A ruleset does not live in the repository, and a required check that has never run once would refuse every release pull request until it did.

Not tested: the status write has never run in Actions. The next push to main exercises it, and #531 has already been merged as 0.9.0, so the next run should write a passing status rather than a failing one.

## Summary by CodeRabbit

* **Chores**
  * Improved automated release validation by reporting whether a release version change passes the configured checks.
  * Release pull requests now display a clear success or failure status before invalid version changes are rejected.
  * No changes were made to product functionality or the end-user experience.


TurtIeSocks:
@coderabbitai full review

TurtIeSocks:
@coderabbitai full review

PR #545, opened 2026-09-21
perf(cli): stop bark allocating twice per bus event

### LINKED ISSUE #520: bark allocates twice per event for answers that are fixed at compile time
Two allocations on bark's per-event path.

- `wire_spelling` round-trips an enum through `serde_json::to_value` to get its wire name (`crates/shep-cli/src/dog/bark/rules.rs:126`), allocating a `Value` on every `on_event`. The spelling is fixed at compile time, so a `match` returning `&'static str` answers the same question with no allocation.
- `self.subjects.entry(subject.to_owned())` (`crates/shep-cli/src/dog/bark/rules.rs:299` and `:337`) allocates a `String` for the key whether or not the subject is already in the map, which on a steady flock is every time. The usual shape is a lookup first, inserting the owned key only on a miss.

Neither is a bottleneck. Both are on the path an event takes through the dog, and both fixes are smaller than what they replace.

ump45nose:
## Summary

Two allocations on bark's per-event path, both in `crates/shep-cli/src/dog/bark/rules.rs`. Neither is a bottleneck; both are on the path an event takes through the dog, and both fixes are smaller than what they replace.

## 1. `wire_spelling` round-tripped the enum through serde

```rust
fn wire_spelling(kind: ProcessEventKind) -> String {
    serde_json::to_value(kind)
        .ok()
        .and_then(|value| value.as_str().map(str::to_owned))
        .unwrap_or_default()
}
```

A `Value` allocated per bus event to answer a question fixed at compile time. Now a `match` returning `&'static str`.

`ProcessEventKind` is `#[non_exhaustive]`, so a wildcard arm is forced. It maps to `"unrecognized"`, which is exactly what `#[serde(other)]` gives `Unrecognized` — i.e. exactly what the round-trip returned for the one variant the wildcard can actually catch today. The spelling a rule compares against is unchanged for every variant this build knows.

**The risk this introduces, and how it is covered.** A variant added upstream gets no arm, falls into the wildcard, and a rule naming it would compare against `"unrecognized"` and never fire — silently, days later. That is why there is a new test:

```rust
fn every_wire_spelling_matches_what_serde_emits()
```

which pins all eleven spellings back against `serde_json::to_value`. I checked it bites by changing `"reload_abandoned"` to `"reload_abandonded"` and watching it fail with `left: "reload_abandonded" / right: "reload_abandoned"`, then restoring. The doc comment on the function names the same obligation.

## 2. `entry(subject.to_owned())` evaluated the key before the lookup

`BTreeMap::entry` takes its key by value, so `subject.to_owned()` ran on *every* call, hit or miss. On a steady flock the subject is already in the map, so every event allocated a `String` it then dropped. Both call sites (`fire` at the old `:299`, `restart_window_crossed` at the old `:337`) now go through one helper:

```rust
fn subject_state<'a>(subjects: &'a mut BTreeMap<String, SubjectState>, subject: &str) -> &'a mut SubjectState {
    if !subjects.contains_key(subject) {
        subjects.insert(subject.to_owned(), SubjectState::default());
    }
    subjects.get_mut(subject).expect("subject key is present: either it was, or the insert above added it")
}
```

Hot path is now `contains_key` + `get_mut` and zero allocations; the owned key is built only on a miss, once per subject per bark run.

Two notes on the shape, since both are non-obvious:

- **`contains_key` first, not a `get_mut` match.** The match form (`match subjects.get_mut(subject) { Some(s) => s, None => subjects.entry(...) }`) does not compile: the reference returned from the lookup has to outlive the insert that follows it, so the borrow of `*subjects` spans the whole match and the `entry` call is a second mutable borrow. E0499. That is also why the early-return `if let` form fails. `contains_key` ends its borrow immediately.
- **The `expect` is unreachable.** The key is present either because it was, or because the line above inserted it. `clippy::expect_used` is a restriction lint and the gate does not enable it, so `-D warnings` is clean.

## Testing

- **`SDKROOT` was pinned to `MacOSX26.sdk`.** The default CLT SDK on this machine (`MacOSX27.0`) has `libSystem.B.tbd` entries for `arm64e.x1-macos` / `arm64e.x1-maccatalyst`, which the local `ld` rejects with `tapi error: malformed file`. Nothing links without the pin. This is a local toolchain problem, not a repo one.
- **I diffed every failure against a pristine clone of `main` at `550b7d1c`.** That matters here, because this machine has a real shepherd 0.1.8 running, which the daemon tests notice.

| gate | result |
| --- | --- |
| `cargo clippy --workspace --all-targets --all-features -- -D warnings` | **10 findings, all pre-existing** — see below |
| `cargo test --workspace --all-features --no-fail-fast` | **8 failures, all pre-existing** — see below |

### The 10 clippy findings are not mine

I ran clippy on the pristine `main` clone and diffed the finding lists: **byte-identical**. All ten come from the local stable being newer than what CI's stable leg currently is — `chunks-exact-to-as-chunks` and the stricter `op_ref` did not exist when this tree was last green. None is in `rules.rs`. My branch adds zero findings.

### The 8 test failures are not mine

Seven in `shep-cli --lib` (`commands::hook::tests::*`) and one in `shep-daemon --lib` (`probes::os::tests::tcp_probe_against_a_non_routable_address_gives_up_within_the_budget`). I confirmed the identical set fails on pristine `main`, with identical counts. The hook tests fail with `error[daemon_unreachable]: no shepherd is running` and `error[version_skew]: this shep is 0.9.0, the running shepherd is 0.1.8`; the probe test is a real-network timing test. Nothing to do with this change.

Against that baseline the numbers are:

- `shep-cli --lib`: **2189 passed** (baseline 2187 — the two extra are my new tests), 3 ignored, same 7 environmental failures.
- `shep-daemon --lib`: **999 passed** (same as baseline), 1 environmental failure.
- Every other target in the workspace: `ok`, no failures.

### What is directly covered

- `dog::bark::rules::tests` — **33 passed, 0 failed**, including the 28 that were already there.
- `every_wire_spelling_matches_what_serde_emits` — verified to fail on a deliberate typo (see above).
- `a_subject_seen_for_the_first_time_gets_its_debounce_state` — new. Two sheep neither of which is in the map yet: each fires once, each is muted inside its own debounce, and each fires again past it. That is the miss path of `subject_state` being exercised end to end, and it is what would break if the insert were dropped and the `get_mut` returned `None`.

## Not changed

Nothing an operator types or sees. No verb, flag, alias, `shep.toml` key, Flockfile field, exit code, JSON shape or default value moved, so the CLI reference pages and `web/src/pages/docs/` need no regeneration and I did not run the site gate. `trigger_name` two functions up already had this exact shape, which is the pattern I followed.

## AI assistance

Written with Claude. Every line above is mine to explain: the serde round-trip is what the issue named, the hand-listing and its wildcard arm are my answer to the future-variant risk that hand-listing creates, the `contains_key` ordering is what the borrow checker forced, and the `expect` message says what it guarantees.

Closes #520

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved event-name handling for unsupported event types, providing consistent fallback labeling.
  * Corrected subject-specific debounce and restart-window behavior so newly observed subjects are tracked independently.
  * Improved consistency between event labels and serialized output across supported event types.

* **Tests**
  * Added coverage for event-name mappings and independent per-subject debounce state.


TurtIeSocks:
@coderabbitai full review

PR #546, opened 2026-09-21
perf(metrics): escape each sheep's label prefix once per scrape

### LINKED ISSUE #514: The metrics dog escapes the same three labels six times per sheep
The metrics dog's status series rebuilds the same three labels six times per sheep (`crates/shep-cli/src/dog/metrics/exposition.rs:208`):

```rust
for candidate in ALL_STATUSES {
    let candidate_string = candidate.to_string();
    let status_pairs = [("sheep", info.name.as_str()), ("id", id_string.as_str()), ("fold", fold), ("status", candidate_string.as_str())];
    status.push(&labels(&status_pairs), value);
}
```

`labels` escapes and formats all four pairs, but only `status` differs between the six iterations. So a scrape pays six escapes of the sheep name, six of the id and six of the fold per sheep, plus a `to_string` for a status whose spelling is known at compile time.

Fix: build the escaped `sheep`, `id` and `fold` prefix once per sheep, and keep the six status strings beside `ALL_STATUSES` as static text. Each candidate then appends its own suffix.

Magnitude: 6N redundant escapes per scrape, and a scrape is every 15 to 60 seconds. Small. It is the dog's entire job, though, and the fix makes the loop read as what it is.

ump45nose:
## Summary

One change, in `crates/shep-cli/src/dog/metrics/exposition.rs`. `render`'s status loop rebuilt the full `{sheep,id,fold,status}` label list for each of the six candidate statuses, so every scrape re-escaped the sheep name, id and fold six times per sheep and allocated a fresh `String` for each line. Only `status` differs between the six. The shared prefix is now escaped once per sheep, and the six status series close it with the one pair that differs.

Closes #514.

## The change

Before, per candidate status:

```rust
let candidate_string = candidate.to_string();
let status_pairs = [
    ("sheep", info.name.as_str()),
    ("id", id_string.as_str()),
    ("fold", fold),
    ("status", candidate_string.as_str()),
];
status.push(&labels(&status_pairs), value);
```

Now the three shared labels are escaped once into an open prefix, and a `Display` adapter closes it with the pair that differs:

```rust
let sheep_prefix = label_prefix(&sheep_pairs);
...
status.push(
    StatusLabels { prefix: &sheep_prefix, status: status_label(candidate) },
    value,
);
```

`MetricGroup::push` takes `impl fmt::Display` for the label list so the adapter can be handed over without being formatted into an intermediate `String`; the fully-closed and label-less call sites are unchanged in behaviour. `status_label` returns `&'static str`, so the loop also drops the `to_string` per candidate.

**The risk this introduces, and how it is covered.** The status spellings become a hand-written `match` sitting beside `dog_source_label`'s, and a typo in one would silently change a label value. Two guards: the spellings are the ones `ProcStatus`'s own `Display` and wire form use, which the doc comment states, and a new golden test pins the entire exposition — all six status series included — byte for byte against a capture taken from the pre-change `render`. The fixture is a mixed flock (a plain sheep, an idle sheep with no CPU/memory samples, a dog, and a sheep whose name needs all three escapes while in a non-online state), so every label path and all six spellings appear in it. After an intentional format change the golden is regenerated with a documented `BLESS=1` command; that regeneration test is `#[ignore]`d and a no-op without the env var, so a stray run cannot move the pin.

## Testing

| gate | result |
| --- | --- |
| `cargo test --workspace --all-features --no-fail-fast` | **8 failures, all pre-existing** — the same 8 fail on the pristine clone (7 `commands::hook::tests` from the running shepherd 0.1.8, 1 `tcp_probe` timing test). `shep` lib: 2188 passed = 2187 + the new golden test |

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Prometheus metrics output to consistently format labels and correctly escape special characters in sheep names.
  - Preserved stable, byte-for-byte metrics output across platforms.
  - Standardized formatting across sheep, dog, daemon, and host metrics.

- **Tests**
  - Added comprehensive coverage for exported metrics, including status, resource, uptime, process, and restart values.
  - Added validation for escaped label values and expected HELP/TYPE metadata.


ump45nose:
CI caught a real one on Windows, and it was mine: the golden fixture is `include_str!`ed and compared byte for byte, and `text=auto` checks it out CRLF on a `core.autocrlf=true` client — a diff on every line, three Windows jobs red. This repository has hit the trap before (see the `.gitattributes` entries for the flockfile schema and the insta snapshots), so the fix is the established one: the fixture now carries `text eol=lf`, same as its neighbours.

Amended into the one commit rather than stacking a fixup, since the pin and the attribute it needs are one thing. Force-pushed; the Windows matrix is re-running. Nothing about the change under test moved.

TurtIeSocks:
@coderabbitai full review

TurtIeSocks:
@coderabbitai full review

crates/shep-cli/src/dog/metrics/exposition.rs
```diff
@@ -195,6 +236,12 @@ pub fn render(reading: &Reading) -> String {
             ("fold", fold),
         ];
         let sheep_labels = labels(&sheep_pairs);
+        // The open form of the same three labels, escaped once here rather
+        // than once per status below: `labels` escapes and formats all of
+        // them, but only `status` differs between the six candidates, so
+        // the closed form would pay six escapes of the sheep name, six of
+        // the id and six of the fold per sheep on every scrape.
+        let sheep_prefix = label_prefix(&sheep_pairs);
```
TurtIeSocks, line 242: Good split, and `label_prefix` is the right seam. It stops one escape short of the title though.

`labels` is `format!("{}}}", label_prefix(pairs))`, so line 238's `labels(&sheep_pairs)` already escapes the name, the id and the fold, and line 244 escapes the same three again. Two passes per sheep per scrape where the comment above 244 says one.

Deriving the closed form from the prefix gets it to one:

```rust
let sheep_prefix = label_prefix(&sheep_pairs);
let sheep_labels = format!("{sheep_prefix}}}");
```

`sheep_pairs` is a fixed three-element array, so the empty-pairs early return in `labels` is not in play here.

TurtIeSocks: Pushed 71477d12 rather than leave this sitting: qwen flagged the same thing independently, which was enough for me to stop waiting on a reply.

`labels` is `format!("{}}}", label_prefix(pairs))`, so both calls ran the same three values through `escape_label_value`. The closed form now derives from the open one instead of rebuilding it.

Output is byte-identical and that is checked, not assumed: `a_full_flock_renders_exactly_the_bytes_the_baseline_did` passes, and dropping the brace off the derived form fails that test and only that one. Happy to drop the commit if you would rather write it yourself.

TurtIeSocks:
@coderabbitai full review

crates/shep-cli/src/dog/metrics/exposition.rs
```diff
@@ -432,4 +491,44 @@ mod tests {
         assert_eq!(super::escape_label_value("a\nb"), r"a\nb");
         assert_eq!(super::escape_label_value(r#"we"b\x"#), r#"we\"b\\x"#);
     }
+
```
TurtIeSocks, line 494: Two label paths here have no test, and your change moves the escaping they flow through, so they seemed worth closing while you are in the file. Both pre-date the PR, so take them or leave them.

Mutations that stay green on `main`: `"adopted"` to `"adoptedX"`, and the `fold` fallback `""` to `"None"`. Both 2186 passed, 0 failed.

```suggestion
    /// `Adopted` is the only source an operator can produce, and the label
    /// is what a dashboard filters on. Renaming it silently repoints every
    /// query built against it.
    #[test]
    fn an_adopted_dog_renders_the_adopted_source_label() {
        let mut dog = sample_info("shep-discord");
        dog.dog = Some(DogSource::Adopted {
            path: "/opt/bin/shep-discord".to_string(),
        });
        let text = render(&Reading {
            flock: vec![dog],
            ..reading()
        });
        assert!(text.contains(r#"source="adopted""#), "{text}");
    }

    /// A sheep in no fold renders an empty label value, not the word
    /// `None` that an `Option`'s own formatting would have produced.
    #[test]
    fn a_sheep_in_no_fold_renders_an_empty_fold_label() {
        let mut info = sample_info("web");
        info.fold = None;
        let text = render(&Reading {
            flock: vec![info],
            ..reading()
        });
        assert!(text.contains("fold=\"\""), "{text}");
        assert!(!text.contains("fold=\"None\""), "{text}");
    }
```

With these, both mutations fail. Verified at `71477d12`.

TurtIeSocks:
@coderabbitai full review

crates/shep-cli/src/dog/metrics/exposition.rs
TurtIeSocks: Both of these gaps are on main already, so they are not yours to fix. Tracked in #576 along with the two tests. Nothing needed on this PR, and thanks for the patience while we sorted out which side of the line they fell on.

crates/shep-cli/src/dog/metrics/exposition.rs
```diff
@@ -432,4 +491,44 @@ mod tests {
...
+    /// The [`Reading`] [`GOLDEN_EXPOSITION`] was captured from: one plain
+    /// sheep, one idle sheep with no CPU or memory sample, one dog, and
+    /// one sheep whose name needs all three escapes while in a non-online
+    /// state, so every label path and all six status spellings appear in
+    /// it.
+    fn golden_reading() -> Reading {
+        let mut dog = sample_info("bark");
+        dog.dog = Some(DogSource::BuiltIn);
+        let mut odd = sample_info(r#"we"b\x"#);
```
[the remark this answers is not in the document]
TurtIeSocks: CodeRabbit is right about the fact. The doc above `golden_reading` says "one sheep whose name needs all three escapes" and `we"b\x` has a quote and a backslash, no newline.

Tried the fix rather than guessing at it. Adding the newline works and the re-bless is mechanical:

```rust
let mut odd = sample_info("we\"b\\x\n");
```

```bash
BLESS=1 cargo test -p shep --lib --all-features bless_exposition_golden -- --ignored
```

Ten fixture lines change, each one the same sheep gaining `\n` at the end of its escaped name, and the seven exposition tests pass.

Worth saying what it does and does not buy, because the wording overstates it a little. A newline bug is already caught: `escape_label_value_handles_backslash_quote_and_newline` covers that directly, and `labels` and `label_prefix` both route through the same `escape_label_value`, so there is no second escaper to get it wrong. What the newline adds is the rendered path. Deleting the `'\n'` arm with the newline in the fixture fails `a_full_flock_renders_exactly_the_bytes_the_baseline_did` as well as the unit test; without it that golden test provably cannot fail on a newline, which is the gap between what the comment claims and what it checks.

Either direction closes it. The newline is the better one, since the golden exists to hold every label path in one place.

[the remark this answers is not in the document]
TurtIeSocks: Agreed, and the reason is better than a doc mismatch. The fixture covers the escape a real name can never carry and skips the one it can.

`normalize_with_home` refuses a name containing `/`, `\` or `:` (crates/shep-core/src/config/normalize/resolved_app.rs:150). So the backslash in `we"b\x` is unreachable for any sheep the daemon ever sees, and the newline the name omits is reachable, since nothing filters it and a TOML string can express it.

Nothing else validates a name: pm2 import refuses anything that does not normalize (`convert.rs`), dog specs call `normalize`, and the daemon only ever holds a `ResolvedApp`.

So adding the newline is the right half of your either/or. It is the one escape the reusable-prefix path can meet in production and the only one currently untested through it.

PR #547, opened 2026-09-21
perf(barks): answer tail requests from a window of the ring's end

### LINKED ISSUE #513: list_barks reads the entire alert history to return the last twenty
`list_barks` returns the last `tail` alerts, and gets them by reading and deserialising the entire history first (`crates/shep-cli/src/whistle/read.rs:169`):

```rust
let mut history = barks::read(&self.paths.barks)?;
let keep_from = history.len().saturating_sub(limit);
history.drain(..keep_from);
```

`$SHEP_HOME/barks.jsonl` is append-only, so the cost of this tool call grows for the life of the installation while the answer stays capped at `MAX_TAIL`. A host that has been alerting for a year pays for a year of alerts to return the last twenty.

Fix: read a bounded chunk from the end of the file and split it into records, growing the window only if it held fewer than `limit` newlines. The bark ring's own cap does not help here, since `barks.jsonl` is the history file rather than the ring.

Found by Comrade Q's performance sweep on 2026-09-19, verified against the code. Not profiled, and on a small history it is invisible, which is exactly why it would go unnoticed until it is not.

ump45nose:
## Summary

`barks.jsonl` is append-only for the life of the installation, and the two callers that want only the recent records read and deserialized every one of them first: `list_barks` in the whistle, and `shep barks --tail N`. A host that had been alerting for a year paid for a year of alerts to be told the last twenty. Both now answer from a window of the file's end.

Closes #513.

## The change

`barks::read_last(path, limit)` in `crates/shep-core/src/barks.rs`, beside `read`:

- A byte window from the end of the file — 64 KiB first, **doubling while the window holds fewer parseable records than were asked for**. The growth counts parseable records, not newlines, so a ring carrying a dead writer's fragment or a future shep's well-formed record still grows until it holds the real thing; a window that starts mid-record drops the fragment rather than offering it to the parser, and a sequence it cuts mid-character is replaced rather than refused, since this file is read during an incident.
- The same answer `read` plus a drain gives: a missing file is an empty ring, an unparseable line is skipped. Total read is bounded by roughly twice the answer's size, not the file's.

Two call sites switched:

- `list_barks` (`crates/shep-cli/src/whistle/read.rs`) — the one the issue names.
- `shep barks --tail N` (`crates/shep-cli/src/commands/dogs/barks.rs`) — the same shape, same fix. `shep barks` without `--tail` still reads the whole ring, which is what it prints.

**The risk this introduces, and how it is covered.** The window logic is new I/O code with real edge cases — growth, the boundary discard, the missing-file and zero-limit paths — and the failure mode that matters is silent: returning the wrong records during an incident. So the equivalence is pinned rather than assumed:

```rust
fn read_last_answers_what_a_full_read_would_for_every_limit()
```

walks every limit from 0 to count+2 over a fixture carrying a dead writer's fragment and a future shep's well-formed record, with records padded past the first window so the growth path is in range too, and asserts `read_last` answers exactly what `read` plus a drain answers. Plus targeted tests for the newest-first order, window growth past a record spanning the first cut, a missing file, and a zero limit.

Checked the tests bite, twice: truncating instead of draining reds three of them, and seeking only when the window starts mid-file reds two (the single-window one stays green, which is the discrimination you want).

## Testing

| gate | result |
| --- | --- |
| `cargo test --workspace --all-features --no-fail-fast` | **8 failures, all pre-existing** — the same 8 fail on the pristine clone (7 `commands::hook::tests` from the running shepherd 0.1.8, 1 `tcp_probe` timing test). `shep-core`: 542 passed = 537 + the 5 new tests |

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed the `barks` command so `--tail N` consistently returns the requested number of most recent records.
  - Updated bark listings to honor trailing-record limits and preserve oldest-first ordering.

- **Performance**
  - Improved tail retrieval by reading only the necessary portion of large history files, reducing memory use and processing time.
  - Malformed records continue to be skipped without preventing valid recent records from being displayed.


TurtIeSocks:
@coderabbitai full review

crates/shep-core/src/barks.rs
```diff
@@ -160,6 +171,72 @@ pub fn read(path: &Path) -> Result<Vec<Bark>, BarkError> {
...
+    let mut bytes = Vec::new();
+    file.read_to_end(&mut bytes)?;
+    let text = if window < len {
+        match bytes.iter().position(|&b| b == b'\n') {
+            Some(newline) => String::from_utf8_lossy(&bytes[newline + 1..]),
+            None => return Ok(Vec::new()),
+        }
+    } else {
+        String::from_utf8_lossy(&bytes)
+    };
```
TurtIeSocks, line 233: CodeRabbit is right here, and it is worth fixing rather than waving through: when `len - window` lands immediately after a newline, the window already starts at a record boundary, but the first-newline scan treats that record's own terminator as the end of a fragment and drops a whole record.

`read_last` hides it by doubling, so the answer stays correct and you pay one extra read. On a PR whose point is fewer reads, that seemed worth closing.

Reading one byte before the window tells the two cases apart:

```suggestion
    let start = len - window;
    // One byte before the window when there is one. A window starting
    // exactly at a record boundary is otherwise indistinguishable from one
    // starting mid-record, and the whole first record gets dropped as if it
    // were a tail.
    let probe = u64::from(start > 0);
    file.seek(SeekFrom::Start(start - probe))?;
    let mut bytes = Vec::new();
    file.read_to_end(&mut bytes)?;
    let text = if start > 0 {
        match bytes.first() {
            Some(b'\n') => String::from_utf8_lossy(&bytes[1..]),
            _ => match bytes.iter().position(|&b| b == b'\n') {
                Some(newline) => String::from_utf8_lossy(&bytes[newline + 1..]),
                None => return Ok(Vec::new()),
            },
        }
    } else {
        String::from_utf8_lossy(&bytes)
    };
```

Ran it in a worktree at `5f2ec3f1`: `cargo test -p shep-core --lib --all-features barks` gives 11 passed, 0 failed, including `read_last_answers_what_a_full_read_would_for_every_limit`.

No test fails without it, since the doubling compensates, so a test pinning the boundary would be a separate call.

crates/shep-core/src/barks.rs
```diff
@@ -23,6 +27,13 @@ use crate::file_lock::FileLock;
 /// Cap the ring keeps itself under when nobody configured one.
 pub const DEFAULT_MAX_BYTES: u64 = 1024 * 1024;
 
+/// The first window [`read_last`] reads from the end of the ring, and the
+/// factor it grows by while the window holds fewer records than were
+/// asked for. A record is a few hundred bytes, so this covers a typical
+/// tail request in one read; the growth is for a caller that asked for
+/// more, or a ring of unusually large records.
+const TAIL_WINDOW_BYTES: u64 = 64 * 1024;
```
TurtIeSocks, line 35: Half of this one I would push back on. IR-26 wants the unit conversion, and that part is fair, so here it is:

```suggestion
/// The first window [`read_last`] reads from the end of the ring, and the
/// factor it grows by while the window holds fewer records than were
/// asked for. 64 KiB. A record is a few hundred bytes, so this covers a
/// typical tail request in one read; the growth is for a caller that
/// asked for more, or a ring of unusually large records.
const TAIL_WINDOW_BYTES: u64 = 64 * 1024;
```

The benchmark basis is the weaker half of the ask. IR-26 exists for a threshold where being wrong costs something you cannot see, and the doubling loop right below makes this one self-correcting: too small and you pay one extra read, which `read_last_grows_the_window_until_it_holds_enough_records` already covers. Worth saying that back rather than inventing a measurement to satisfy the rule.

crates/shep-core/src/barks.rs
```diff
@@ -294,6 +371,108 @@ mod tests {
         assert_eq!(read(&dir.path().join("nothing.jsonl")).unwrap(), vec![]);
     }
 
+    #[test]
+    fn read_last_on_a_missing_file_is_no_barks_rather_than_a_failure() {
```
TurtIeSocks, line 375: Test for the boundary fix above. Without it: `left: ["s2"], right: ["s1", "s2"]`. With it, the barks suite is 12 passed, 0 failed.

```suggestion
    /// A window that begins exactly at a record boundary holds that record
    /// whole, so the scan that drops a partial record's tail must not eat
    /// it.
    #[test]
    fn a_window_starting_on_a_boundary_keeps_its_first_record() {
        let dir = tempfile::tempdir().unwrap();
        let path = dir.path().join("barks.jsonl");
        for i in 0..3u64 {
            append(&path, &bark_for(&format!("s{i}"), i), DEFAULT_MAX_BYTES).unwrap();
        }
        let bytes = std::fs::read(&path).unwrap();
        let len = bytes.len() as u64;
        let first = bytes.iter().position(|&b| b == b'\n').unwrap() as u64;
        let window = len - (first + 1);
        let mut file = File::open(&path).unwrap();
        let barks = read_window(&mut file, len, window).unwrap();
        assert_eq!(
            barks.iter().map(|b| b.subject.as_str()).collect::<Vec<_>>(),
            ["s1", "s2"]
        );
    }

    #[test]
    fn read_last_on_a_missing_file_is_no_barks_rather_than_a_failure() {
```

crates/shep-core/src/barks.rs
TurtIeSocks: Measured it, so the benchmark half can be satisfied rather than argued down.

A bark carrying one sink outcome serialises to 182 bytes. A shepherd-written one with no sinks is 96. So 64 KiB holds roughly 360 of the first kind or 680 of the second, and it is 6 percent of a 1 MiB `DEFAULT_MAX_BYTES` ring, four doublings from covering a full one.

That gives the comment a number instead of "a few hundred bytes", which is the part IR-26 is actually after. Your unit conversion plus those figures closes both halves, and nobody has to decide whether the rule admits an exception.

crates/shep-core/src/barks.rs
```diff
@@ -160,6 +171,72 @@ pub fn read(path: &Path) -> Result<Vec<Bark>, BarkError> {
...
+        if barks.len() >= limit || window >= len {
+            let keep_from = barks.len().saturating_sub(limit);
+            barks.drain(..keep_from);
+            return Ok(barks);
+        }
+        // Doubling rather than a fixed step: the window that finally
+        // suffices is within a factor of two of this one, so the total
+        // read is bounded by twice the answer's size rather than by the
+        // file's.
+        window = (window * 2).min(len);
```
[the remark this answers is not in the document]
TurtIeSocks: Unreachable, and still worth one word.

`window` starts at `min(TAIL_WINDOW_BYTES, len)` and the loop only continues while `window < len`, so `window * 2` overflows only when `len` exceeds 2^63, about 9.2 exabytes. `DEFAULT_MAX_BYTES` is 1 MiB and no filesystem anyone runs this on approaches that ceiling.

The mechanism is right though. Wrapping to zero gives `read_window(.., 0)`, which seeks to `len`, reads nothing and returns empty, so `barks.len()` stays 0 and `window` stays 0 and the loop never ends. `saturating_mul(2)` costs nothing and saves the next reader from deriving all of that again.

crates/shep-core/src/barks.rs
TurtIeSocks: Confirmed by the mechanism, not only by the failing test.

Three records s0 s1 s2, and a window starting exactly at s1's first byte. `read_window` reads `s1\ns2\n`, the first-newline scan finds s1's own terminator, and everything before it is dropped as a fragment tail. Text becomes `s2\n`.

That is precisely the `left: ["s2"], right: ["s1", "s2"]` you measured. Reading the byte before the window is the right discriminator, since a boundary start and a mid-record start are otherwise identical from inside the window.

PR #548, opened 2026-09-21
perf(metrics): sample the host off the tokio worker

### LINKED ISSUE #522: The metrics dog walks the process table on the async scrape path
`sample_host` builds a `sysinfo::System` with process refresh on (`crates/shep-cli/src/dog/metrics/mod.rs:113`), which walks the whole process table. It is called from `handle_connection`, an `async fn` (`crates/shep-cli/src/dog/metrics/mod.rs:205`, call at `:253`), between two awaits.

So the walk runs on a tokio worker thread with nothing yielding. For the length of that walk the worker cannot serve any other task, including the dog's own shepherd channel.

Fix: `tokio::task::spawn_blocking` around the sample, awaiting its result. The function is already synchronous and takes no borrow from the async context, so it moves as it stands.

Magnitude: once per Prometheus scrape, typically every 15 to 60 seconds, and the walk is bounded by the host's process count. The comment above it already narrows the refresh as far as it can, including turning tasks off so Linux does not count every thread. What it cannot do is stop the walk blocking the runtime.

`whistle::read` also calls this, per its doc comment, so whichever fix lands should suit both callers.

ump45nose:
## Summary

`sample_host` walks the whole process table, and `handle_connection` called it inline between two awaits: for the length of the walk, the tokio worker serving that scrape could serve no other task — the dog's own shepherd channel included. Once per scrape is often and the walk is bounded by the host's process count, so the stall is short, but it is a stall on the one thread the dog needs responsive. The walk now runs on the blocking pool.

Closes #522.

## The change

One helper in `crates/shep-cli/src/dog/metrics/mod.rs`, beside `sample_host`:

```rust
pub(crate) async fn sample_host_off_worker() -> Option<HostReading> {
    tokio::task::spawn_blocking(sample_host)
        .await
        .unwrap_or_default()
}
```

Both async callers go through it:

- `handle_connection` — the scrape path the issue names.
- the whistle's `get_metrics` (`crates/shep-cli/src/whistle/read.rs`) — the same shape, per the issue's note that whichever fix lands should suit both callers. This tool had no test at all before; it has one now.

`sample_host` borrows nothing from the async context, so it moves as it stands. A panic inside the pool reads as `None`: `sample_host` is sysinfo calls that do not panic, so this cannot happen, and answering without the host block beats taking the caller down with it. The doc comment on the helper says so.

**The risk this introduces, and how it is covered.** The failure mode is silent: break the boundary plumbing and every existing assertion stays green while the host series quietly vanish from every scrape. So the boundary is pinned at both ends — `a_scrape_carries_the_host_sample` asserts the host series survive into the scrape response, `get_metrics_carries_the_host_sample` asserts the host row survives into the tool's reply. Checked both bite by making the helper answer `None` unconditionally: two red.

**What this does not test.** The property itself — that the walk no longer blocks a worker — cannot be asserted without a timing-sensitive test, and a flaky test is worse than none. What is pinned is that the sample still arrives, on every platform the gate runs on.

## Testing

| gate | result |
| --- | --- |
| `cargo test --workspace --all-features --no-fail-fast` | **8 failures, all pre-existing** — the same 8 fail on the pristine clone (7 `commands::hook::tests` from the running shepherd 0.1.8, 1 `tcp_probe` timing test). Everything else green: 4043 passed |

**Not tested:** Linux and Windows, and the non-blocking property itself. CI's half of the gate.

## Summary by CodeRabbit

* **Performance**
  * Improved metrics collection responsiveness by moving host data sampling off the asynchronous execution path.
  * Preserved host memory and uptime metrics during scrapes.
  * Maintained existing flock retrieval behavior for each metrics request.


TurtIeSocks:
@coderabbitai full review

TurtIeSocks:
@coderabbitai full review

PR #550, opened 2026-09-21
perf(cli): deliver bark's sinks together, not in turn

### LINKED ISSUE #519: bark delivers to its sinks in turn, so one dead endpoint delays the rest
`deliver_and_record` awaits each sink in turn (`crates/shep-cli/src/dog/bark/firing_delivery.rs:77`):

```rust
for name in &firing.sinks {
    ... super::sinks::deliver(sink, &bark, delivery.sink_timeout).await
```

So one firing to three sinks waits for the first to finish before it starts the second. With a per-sink timeout, a dead endpoint costs the whole timeout before the healthy sinks hear anything, and three dead ones cost three timeouts in a row.

This is a latency property rather than a throughput one: an alert that matters is delayed by a sink that does not.

Fix: drive the sinks concurrently, collecting the outcomes, since each one already carries its own timeout and its own `SinkOutcome`. The ordering of `outcomes` has to stay stable if anything downstream reads it positionally, which is worth checking before merging.

ump45nose:
## Summary

`deliver_and_record` awaited each sink in turn, so one firing to three sinks waited for the first to finish before it started the second. With a per-sink timeout, a dead endpoint cost the whole timeout before the healthy sinks heard anything, and three dead ones cost three timeouts in a row. This is a latency property rather than a throughput one: an alert that matters is delayed by a sink that does not. The sinks are now driven together.

Closes #519.

## The change

One function, `crates/shep-cli/src/dog/bark/firing_delivery.rs`: the per-sink loop becomes a `join_all` over one future per sink, polled together on the delivery's own task. `futures-util` was already in this crate's tree for its streams, so this adds no dependency.

Three things the issue asked to have checked before merging, and what the check found:

- **Outcome order.** `join_all` hands the outcomes back in `firing.sinks`' order, which is the order the trail has always recorded them in. I also checked what reads them downstream: production maps `bark.sinks` with `iter()` into `BarkRow` — no positional semantics anywhere; the one positional read in the tree is a single-sink test assertion.
- **Per-sink timeouts.** Each delivery still carries its own `Delivery::sink_timeout` and its own `SinkOutcome`; a dead endpoint still costs exactly its own timeout, it just no longer costs anyone else's.
- **The append lock.** Unchanged: it still covers only the `barks::append` read-modify-rename, which is what several concurrent firings contend on.

**The risk this introduces, and how it is covered.** The failure mode is a sink that stops being delivered to at all, or outcomes that reorder. The new test pins the property as an order rather than a duration, which is what the existing `slow_sink` helper's doc asks for:

```rust
async fn a_dead_sink_does_not_delay_the_live_ones_delivery()
```

A firing naming a hanging sink first must still have the live sink's request arrive while the dead one hangs — in turn, the live sink is not contacted until the dead sink's timeout has already passed. The recorded outcomes must come back in the firing's order, with the dead one's timeout as its own error and the live one clean. Checked it bites by restoring the in-turn loop: the test fails with the live sink uncontacted inside the window.

## Testing

| gate | result |
| --- | --- |
| `cargo test --workspace --all-features --no-fail-fast` | **8 failures, all pre-existing** — the failing set is identical to the pristine clone's (7 `commands::hook::tests` from the running shepherd 0.1.8, 1 `tcp_probe` timing test); everything else green, 4042 passed |

## Summary by CodeRabbit

- **Improvements**
  - Deliveries to configured destinations now run concurrently, improving responsiveness when some destinations are slow or unreachable.
  - Delivery results continue to be recorded in firing order for consistent reporting.


TurtIeSocks:
@coderabbitai full review

TurtIeSocks:
@coderabbitai full review

crates/shep-cli/src/dog/bark/firing_delivery.rs
```diff
@@ -182,4 +190,83 @@ mod tests {
...
+            .await
+            .expect(
+                "the live sink must hear about the alert while the dead one still hangs; \
+                 driven in turn it would have waited out the dead sink's timeout first",
+            )
+            .expect("the live sink must have captured its request");
+
+        // The delivery then finishes on the dead sink's own timeout, with
+        // both outcomes in the firing's order.
+        recorded.await.expect("the delivery task must not panic");
```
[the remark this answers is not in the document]
TurtIeSocks: Fixed in 3886dbd2, five seconds against the 500ms `sink_timeout` the test sets. Confirmed it actually fires rather than just sitting there: cut to 1ms the test fails in 0.02s with `Elapsed(())`.

The same shape is in `a_bark_is_recorded_even_when_every_sink_refuses_it` lower down, an unbounded await on `deliver_and_record` itself. That one is on main unchanged so I left it out of this PR and filed it separately.

TurtIeSocks:
@coderabbitai review

TurtIeSocks:
@coderabbitai review

TurtIeSocks:
@coderabbitai review

crates/shep-cli/src/dog/bark/firing_delivery.rs
```diff
@@ -63,6 +64,14 @@ pub(super) fn spawn_firings(firings: Vec<Firing>, delivery: &Delivery) {
...
+/// timeout runs beside the healthy sinks' deliveries instead of ahead of
+/// them, so one unreachable webhook no longer holds up an alert the
+/// others could have carried. Each delivery keeps its own
+/// [`Delivery::sink_timeout`] and its own [`SinkOutcome`], and `join_all`
+/// hands the outcomes back in `firing.sinks`' order, which is the order
+/// the trail has always recorded them in.
+///
 /// After delivery, since a [`Firing`]'s [`shep_core::barks::Bark::sinks`]
 /// is empty until each sink has been tried. Written even when every sink
 /// refused it: the local trail is what an operator reads when the page
```
TurtIeSocks, line 77: This paragraph lost its sentence. It leans on "writes the resulting `Bark` to `barks_path`" back on line 64, and the new paragraph above pushed the two seven lines apart, so "After delivery, since..." now reads as a fragment with nothing to attach to.

The text is older than your change, but the change is what made it unreadable. Giving it its own subject fixes it and touches nothing else:

```suggestion
/// The bark is written after delivery, since a [`Firing`]'s
/// [`shep_core::barks::Bark::sinks`] is empty until each sink has been
/// tried. It is written even when every sink refused it: the local trail
/// is what an operator reads when the page
```

TurtIeSocks: Checked this against main and the fragment is already there, so it is ours rather than yours. Tracked in #575. Nothing needed on this PR.

PR #551, opened 2026-09-21
perf(daemon): four of the per-tick repeated-work findings

### LINKED ISSUE #521: Five small pieces of repeated work in the daemon's per-tick and listing paths
Four small ones in the daemon, each verified, none individually worth an issue.

- `total_over` allocates a `HashSet` and pays a hash insert per node walked, to guard against a parent-link cycle the kernel cannot produce, since a pid has one parent and the graph is a forest (`crates/shep-daemon/src/limits/sample.rs:237`). This is per armed root per tick. Removing the guard needs a look at whether any test fixture builds a cycle on purpose.
- In the same build loop, `children_of` uses `HashMap::new()` while its two siblings use `with_capacity(table.len())` (`crates/shep-daemon/src/limits/sample.rs:200`), so one of the three rehashes its way up per tick and the other two do not.
- `names.contains(&info.name)` scans the restored-apps list for every row of the flock listing (`crates/shep-daemon/src/rpc/dispatch.rs:341`), which is flock times restored string comparisons on a `Muster` call. A set built once makes it one pass.
- `value.as_str().to_owned()` copies a `String` that `entries.into_iter()` already yielded owned (`crates/shep-daemon/src/rpc/dispatch.rs:695`), doubling the allocations when building the values map for a secrets push.

And one more, adjacent: `respawn` clones `ResolvedApp` at `crates/shep-daemon/src/supervisor/actor_lifecycle.rs:103` and clones it again at `:131` to hand it to `spawn_sheep_task`, so a crash loop pays two full config copies per respawn where one clone plus a move would do.

Found by Comrade Q's performance sweep on 2026-09-19, all verified against the code. None measured.

ump45nose:
## Summary

Four of the issue's five findings, as three focused commits — one per thing, so each reverts alone. The fifth is deliberately not in this PR; the reason is below, and it is the one finding I would argue with.

Closes #521.

## What landed

**`perf(daemon): size the children map like its two siblings`** — `TreeIndex::build` (`limits/sample.rs`) sized two of its three maps with `with_capacity(table.len())` and started `children_of` from empty, so one of the three rehashed its way up per build — once per armed root per tick on the polling enforcer's path.

**`perf(daemon): two repeated-work findings in the rpc paths`** (`rpc/dispatch.rs`, `protocol/request/redacted.rs`):

- The Muster arm filtered the flock listing with `names.contains` per row — flock times restored string comparisons on every Muster. A set built once from the names makes it one pass.
- The secrets-push arm built its values map with `value.as_str().to_owned()`, copying a `String` that `entries.into_iter()` had already yielded owned. `EnvValue::into_string` takes the `String` out instead — the consuming accessor the config-side `EnvValue` already carries for the same reason.

Neither changes an answer: the filter's predicate is the same string equality, the values map holds the same bytes. `misc_verbs`' Muster reply and `push`'s stored entries pin both, and pass unchanged.

**`perf(daemon): hand the resolved app to the spawn instead of cloning it twice`** (`supervisor/actor_lifecycle.rs`) — `respawn` cloned the entry's spec once to read it against the slot, then cloned it again to hand it to `spawn_sheep_task`, so a crash loop paid two full config copies per respawn. The second consumer's only other use was `app.config().listen_timeout`, a `Copy` value now read before the move. The lifecycle suites — respawn, readiness, gated and ungated arms — pass unchanged, which is the whole claim: same process, same handles, one fewer copy.

## The finding I did not take, and why

**`total_over`'s cycle-guard `HashSet` stays.** The issue's premise is that the kernel cannot produce a parent-link cycle, so the guard is dead weight. But the same file family says otherwise in its own comment, on the walk that shares the rationale (`lambs_of`, `limits/stats.rs`):

> the kernel does not produce a cycle in the parent links, but a fixture can and **a torn `/proc` read might**

The sampler reads `/proc` across two instants; a pid recycled between them can present a parent link that closes a loop. The guard turns that into a bounded walk instead of an unbounded one, and it is cheap — the walk is already `O(tree)` with a hash lookup per node. Removing it would also make `lambs_of`'s deliberate cycle fixture (`a_parent_link_cycle_terminates_and_reports_each_pid_once`) inconsistent with its sibling, and that fixture is exactly the "test fixture builds a cycle on purpose" the issue asked to have checked. So: checked, and the guard earns its allocation. If a future change makes the sampler's reads atomic, this is worth another look.

## Testing

| gate | result |
| --- | --- |
| `cargo test --workspace --all-features --no-fail-fast` | **8 failures, all pre-existing** — the failing set is identical to the pristine clone's (7 `commands::hook::tests` from the running shepherd 0.1.8, 1 `tcp_probe` timing test); everything else green, 4041 passed |

Targeted runs before the full suite: `limits` 45 passed, `rpc::` 115 passed, `supervisor` 294 passed.


TurtIeSocks:
Three of the four are clean wins. The `HashSet<&str>` is the big one, and hoisting `listen_timeout` so `app` can move instead of clone is the kind of saving that only shows up on a crash loop.

`d9c6dd99` is mislabelled though. It is `perf(daemon)` and it adds `pub fn into_string` to `EnvValue` in `crates/shep-core/src/protocol/request/redacted.rs`, which `protocol/mod.rs:39` re-exports. That is new public API on a published crate.

release-plz attributes by touched file, so shep-core takes a Performance entry and a patch bump for something additive, which is a minor. `semver_check = true` does not catch it, because additive is not breaking. Same shape as the `BusEvent::topic` patch CLAUDE.md records, without the break.

Split into `feat(core):` for the method and `perf(daemon):` for the dispatch change, and both the scope and the bump come out right.

TurtIeSocks:
@coderabbitai full review

TurtIeSocks:
@coderabbitai full review

crates/shep-daemon/src/limits/sample.rs
```diff
@@ -197,7 +197,10 @@ impl TreeIndex {
         // the whole-machine table per root.
         let mut bytes_by_pid: HashMap<u32, u64> = HashMap::with_capacity(table.len());
         let mut cpu_by_pid: HashMap<u32, u64> = HashMap::with_capacity(table.len());
-        let mut children_of: HashMap<u32, Vec<u32>> = HashMap::new();
+        // Sized like the two maps above rather than from empty: at most one
+        // key per row, and growing from nothing rehashes its way up once
+        // per armed root per tick.
```
TurtIeSocks, line 202: The capacity hint is right, the cost it names is not. `build` runs once per table, so the rehash it avoids is once a tick rather than once per armed root.

The struct doc ten lines up makes exactly that point: build one per table and call `sum_from` per root, because rebuilding per root would multiply the per-tick cost by flock size. As written this comment says the rehash does scale with flock size, which is the thing the type exists to stop.

```suggestion
        // Sized like the two maps above rather than from empty: at most one
        // key per row, and growing from nothing rehashes its way up on every
        // build, which is once a tick.
```

Worth fixing rather than waving through, because the number is what a later reader would reach for if they ever wanted to tune the hint.

TurtIeSocks (commented):
Qwen round over the diff, cut as hunks against the merge base. One finding, and it holds up: a comment that names a cost the type is designed to avoid.

TurtIeSocks:
@coderabbitai full review

TurtIeSocks:
@coderabbitai full review

crates/shep-core/src/protocol/request/redacted.rs
```diff
@@ -80,6 +80,14 @@ impl EnvValue {
...
         &self.0
     }
+
+    /// The value, consuming: the caller already owns the entry, so
+    /// borrowing through [`Self::as_str`] and copying would double the
+    /// allocation for a `String` already in hand.
+    #[must_use]
+    pub fn into_string(self) -> String {
+        self.0
+    }
```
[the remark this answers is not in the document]
TurtIeSocks: Confirmed, and it is genuinely reachable. `mod redacted` is private, but `request/mod.rs:20` re-exports with `pub use redacted::{DogSectionToml, EnvValue}` and `protocol/mod.rs:32` is `pub mod request`. So the new method lands as `shep_core::protocol::request::EnvValue::into_string`.

`d9c6dd99` touches both crates, so it is shep-core's commit too, and `perf` files the entry under the daemon at patch level. shep-core gains a public method with nothing in its own changelog saying so.

Neither existing guard catches this. `commits.yml` checks that a subject parses, not that the type matches the change, and cargo-semver-checks stays quiet because an addition breaks nothing. This is the same shape that sent a changed `BusEvent::topic` signature out as an empty-changelog patch in 0.2.1.

Splitting the accessor into its own `feat(core):` commit is the fix.

crates/shep-daemon/src/limits/sample.rs
TurtIeSocks: Checked the call sites and the comment is backwards, so this is worth taking.

`limits/mod.rs:117` calls `TreeIndex::build(&table)` once, outside the loop, and `sum_from` runs per armed id inside `guard.retain` at :132. So the rehash `with_capacity` avoids happens once a tick, not once per armed root.

The comment eight lines above the build says it plainly: "One sample pass and one index build serve every armed sheep. Without the hoist a tick costs O(flock times table size)." The inline comment describes the cost the hoist exists to remove, which is the one thing a later reader tuning the hint must not believe.

PR #554, opened 2026-09-21
fix(cli): refuse a config value the field's own schema does not take

TurtIeSocks:
The VALIDATION panel printed `500ms, 2s, 5m, 1h` beside a dog's `max_age` and the pane filed `1d` anyway. `shep-log-rotate` then failed `Config::from_toml` on every tick, rotated nothing, and went on reporting online. Only its own log said so.

- `apply_typing` checks the buffer against the field's own schema before filing it
- shep-core's two string grammars go through their `FromStr`, resolved off the `$ref` name, so a dog's `Option<UpDuration>` reaches it too
- any other string with a `pattern` is matched against it: the only door a third-party dog's grammar has
- an integer honours the schema's `minimum` and `maximum`, which is what `shep-log-rotate`'s `keep = 0` floor was already asking for
- a refusal keeps the editor open and names the key, the value, and what the field takes. The integer case used to refuse in silence
- a pattern that does not compile refuses nothing, so a dog's broken regex cannot lock an operator out of a field

Why it has to be here: `Request::SetDogConfig` checks the table is TOML and writes it, since the dog owns its own config, and lookout is that request's only sender. A sheep's write is re-validated by the daemon, so for one of those this is the earlier of two gates rather than the only one.

Not public API: `mod lookout` is private in shep-cli.

The other half is the dog's, and it is not in this PR. A dog that cannot read its config retries forever and stays online, which is the anti-pattern `/docs/writing-a-dog` already names for shep's own bark dog. `shep-log-rotate` should exit on it, the way it already exits on a refused handshake.

Gates: `cargo fmt --check`, `cargo clippy --workspace --all-targets --all-features -D warnings`, `cargo test --workspace --all-features`, `RUSTDOCFLAGS="-D warnings" cargo doc --document-private-items`, and `npm ci` + `astro check` + `npm run build` in `web/`. All green.


TurtIeSocks:
Read this for the surface the pattern arm opens, since it compiles a regex a third party wrote.

`regex` is linear-time by construction, so a dog's hostile pattern cannot stall the pane the way a backtracking engine would. And `Regex::new` runs once per commit, not per keystroke: `refusal` has one call site, `pane/edit.rs:433`, and `apply_typing` takes `self.typing` and files. The name reads like a keystroke handler, which is the only reason me checked.

Failing open on a pattern that does not compile is the right call, and `a_pattern_that_does_not_compile_refuses_nothing` pins it.

The integer double parse is fine, and the comment above it explains the `Err` arm better than a reviewer would have asked for.

Nothing blocking from me. Still wants a CodeRabbit read and a qwen round.

TurtIeSocks:
Two minor things in `validation.rs`:

- `table_for`'s Bool arm is unpinned. Gutting it to `FieldKind::Bool => (&[], &[])` leaves 2198/2198 green on `cargo test -p shep --lib --all-features -- --skip ::slow::`. The Integer arm is covered, but only by `dog_pane_wide.snap` carrying "a whole number", not by a `bullets()` test.
- `MemSize` gets an empty refusals list where `UpDuration` gets `DURATION_REFUSALS`. `a_size_field_refuses_a_spelling_shep_does_not_take` already names "10MB" as the natural wrong spelling. Deliberate, or worth a `MEMORY_REFUSALS` table?

Nothing else on the diff. Full lib suite green at 42a108f5.

TurtIeSocks:
@coderabbitai full review

TurtIeSocks:
@coderabbitai full review

crates/shep-cli/src/lookout/validation.rs
```diff
@@ -132,6 +139,141 @@ pub fn bullets(field: &Field) -> Bullets {
...
+        Some(ValueKind::UpDuration) => typed
+            .parse::<UpDuration>()
+            .is_err()
+            .then(|| refuse(&field.key, typed, "a duration", DURATION_FORMS)),
+        Some(ValueKind::MemSize) => typed
+            .parse::<MemSize>()
+            .is_err()
+            .then(|| refuse(&field.key, typed, "a size", MEMORY_FORMS)),
+        None => pattern_refusal(field, typed),
+    }
```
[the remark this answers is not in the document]
TurtIeSocks: Verified and fixed in 690ef839. Both doors were open, and the real code printed these:

```
token is "sk_live_HUNTER2", which it does not take; try it matches ^sk-
api_key is "hunter2", which is not a duration shep accepts; try 500ms, ...
```

`refusal` now resolves one `shown` value ahead of every path, `<set>` for a secret and the buffer otherwise. `typed` is still what gets parsed, so the verdict is unchanged and only the sentence moves.

Worth adding to your reachability note: this was the only place in the pane a secret's value could appear. `begin_typing` seeds a secret's editor empty and the row renders `<set>`, both so it is never drawn.

Exact strings pin the sentence and the `Debug`, plus a control that keeps an ordinary field quoting what was typed. Dropping the redaction fails that test alone.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

crates/shep-cli/src/lookout/field.rs
```diff
@@ -280,6 +314,34 @@ fn strip_nullable<'a>(schema: &'a Value, defs: &'a Map<String, Value>) -> &'a Va
...
+/// mislead.
+fn bounds_of(schema: &Value, defs: &Map<String, Value>) -> Bounds {
+    let schema = resolved(schema, defs);
+    Bounds {
+        pattern: schema
+            .get("pattern")
+            .and_then(Value::as_str)
+            .map(str::to_owned),
+        minimum: schema.get("minimum").and_then(Value::as_i64),
+        maximum: schema.get("maximum").and_then(Value::as_i64),
```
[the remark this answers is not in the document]
TurtIeSocks: Verified and fixed in 690ef839. `minimum: 0.5, maximum: 2.5` produced `Bounds::default()`, and the pane accepted `0` and `3` and filed both.

Bounds are read as `f64` now and normalised to the integer they forbid, up for a minimum and down for a maximum.

`unsatisfiable` covers the rest: a bound past `i64` in the direction that admits nothing, and a range inverted in the schema or by rounding. The other direction bounds nothing, so it stays `None` rather than reject-all. A field with one says so once instead of refusing `1` for being under the floor and `3` for being over the ceiling.

Kept the f64 comparison rather than a checked conversion. It cannot name every `i64` near its ends, so a bound within a few units of `i64::MAX` saturates to it. Documented, and no operator is typing sixteen digits into a config pane.

Restoring `as_i64` fails both bounds tests.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

TurtIeSocks:
Verified 690ef839. The redaction holds: replacing `shown` with `typed` at `validation.rs:204` fails `a_secrets_value_never_reaches_the_sentence_or_the_debug`, with `token is "sk_live_HUNTER2", which it does not take; try it matches ^sk-` against the expected `<set>`. 2201 pass, that one fails, so the test is pinning the thing it claims to.

Two hazards it leaves behind, neither of them live:

- `refuse` at `:300` still formats `{typed}`, and its parameter is named `typed`. Both call sites pass `shown`, so nothing leaks today. It is the one builder in the set whose signature invites the raw value back, where the other three take an explicit `shown` and carry a doc line saying what it is.
- `not_a_whole_number` is `pub(super)`, and `edit.rs:452` calls it with the raw buffer. Unreachable right now, since `integer_refusal` returns `Some` for every buffer that fails to parse, so the early return in `apply_typing` always fires first. The guarantee is a comment rather than a type, on the one function that prints the value.

Separately, `field.rs` goes from 894 on main to 1120 here, so this is the change that takes it past the 1000 line rule.

Both findings from my earlier round still stand at this head: `table_for`'s Bool arm is pinned by nothing, and `MemSize` still gets an empty refusals list.

TurtIeSocks:
Rin's call on the size note above: this is a requirement rather than a follow-up. `field.rs` is under the limit on `main`, so #554 should not be the change that takes it over, and deferring means merging 1120 lines in order to fix them later.

#574 is retargeted to this PR.

crates/shep-cli/src/lookout/field/init.rs
```diff
@@ -0,0 +1,175 @@
...
+pub(super) fn strings(init: Option<&Value>, key: &str) -> Vec<String> {
+    let Some(values) = init.and_then(|i| i.get(key)).and_then(Value::as_array) else {
+        return Vec::new();
+    };
+    let names: Vec<String> = values
+        .iter()
+        .filter_map(Value::as_str)
+        .map(str::to_owned)
+        .collect();
+    if names.len() == values.len() {
```
TurtIeSocks, line 35: `suggestions` and `strings` are the same shape twice: `filter_map(Value::as_str)`, collect, then compare the collected length against the original array. The all or nothing contract lives in two places, so a change to it (trimming, rejecting empties) lands in one and not the other.

Neither negative path has a test. Every case in the module feeds an all-string array, so `["SIGTERM", 42]` returning `None` and `["ok", 7]` returning an empty vec are both unexercised, and the length check is the only thing enforcing either.

TurtIeSocks (commented):
Qwen round over this branch's diff against its merge base, cut as annotated hunks rather than whole files. Eleven findings, ten of which hold up on reading, all in new code rather than in what the split moved.

One dismissed: it called the `Text` to `Suggested` promotion untested in `field/mod.rs`, but `init.rs` asserts `FieldKind::Suggested(vec!["SIGTERM", "SIGINT"])` and a second test pins the non-empty guard. The tests are in the sibling module, not missing.

The two worth doing something about are the `blurb` fallback, which contradicts its own doc, and the empty `enum`, where the `oneOf` sibling already guards what this one does not. The rest are small.

Multiset diff of the split says nothing was lost: twelve line instances differ and every one is a private `fn` that gained `pub(super)`.

crates/shep-cli/src/lookout/field/mod.rs
```diff
@@ -227,178 +243,6 @@ impl FieldSet {
...
-            let field = entry.get("field")?.as_str()?.to_owned();
-            let note = entry.get("note")?.as_str()?.to_owned();
-            Some(Neighbour { field, note })
-        })
-        .collect()
-}
-
 fn field_from(key: &str, schema: &Value, defs: &Map<String, Value>) -> Field {
     let init = schema.get("init");
     let help = init
```
TurtIeSocks, line 248: `or_else` runs on `Option<&Value>`, before `as_str`. A `blurb` that is present but not a string is `Some(_)`, so the fallback never fires and `as_str` then yields `None`, leaving the key.

The doc on `help` says `init.blurb`, else `description`, else the key, so this is a contradiction rather than a judgement call. `help_prefers_the_blurb_then_the_description_then_the_key` passes because all three of its cases use a string blurb or none at all.

```rust
    let help = init
        .and_then(|i| i.get("blurb"))
        .and_then(Value::as_str)
        .or_else(|| schema.get("description").and_then(Value::as_str))
        .map_or_else(|| key.to_owned(), str::to_owned);
```

crates/shep-cli/src/lookout/field/schema.rs
```diff
@@ -0,0 +1,316 @@
...
+            return FieldKind::Choice(names);
+        }
+    }
+    if let Some(values) = schema.get("enum").and_then(Value::as_array) {
+        let names: Vec<String> = values
+            .iter()
+            .filter_map(Value::as_str)
+            .map(str::to_owned)
+            .collect();
+        if names.len() == values.len() {
```
TurtIeSocks, line 106: The `oneOf` arm above guards with `!names.is_empty() && names.len() == consts.len()`. This one drops the first half, so `enum: []` satisfies `0 == 0` and returns `Choice(vec![])`, which renders as a dropdown with no options.

The `enum` branch also has no test at all. `"enum"` appears exactly once in the whole lookout tree and it is the `schema.get` on this line, so every choice test goes through `oneOf`.

crates/shep-cli/src/lookout/field/schema.rs
```diff
@@ -0,0 +1,316 @@
...
+            Some("integer") => FieldKind::List(ListItem::Integer),
+            _ => FieldKind::Opaque,
+        },
+        Some("object")
+            if schema.get("additionalProperties").is_some()
+                && schema.get("properties").is_none() =>
+        {
+            FieldKind::Map
+        }
+        _ => FieldKind::Opaque,
```
TurtIeSocks, line 127: `strip_nullable(resolve(items(schema), defs), defs)` is the two hops `resolved` already holds, and `resolved`'s own doc says it exists to be the one spelling shared by its callers so a reader that followed only one would report a nullable field's bounds as absent. The array items path is the one caller that does not use it.

crates/shep-cli/src/lookout/field/schema.rs
```diff
@@ -0,0 +1,316 @@
...
+        Value::String(s) => Some(s.as_str()),
+        Value::Array(arr) => arr.iter().filter_map(Value::as_str).find(|t| *t != "null"),
+        _ => None,
+    }
+}
+
+pub(super) fn kind_of(schema: &Value, defs: &Map<String, Value>) -> FieldKind {
+    let schema = resolved(schema, defs);
+    if let Some(consts) = schema.get("oneOf").and_then(Value::as_array) {
+        let names: Vec<String> = consts
```
TurtIeSocks, line 91: `consts` holds the `oneOf` arm objects, each a whole schema, and the constants come out into `names` on the next line. `arms` would match `strip_nullable` and `ref_name` and stop the name promising the extracted values.

crates/shep-cli/src/lookout/field/schema.rs
```diff
@@ -0,0 +1,316 @@
...
+
+/// `anyOf: [T, {type: null}]` is `T` with the field optional. Anything
+/// else is left as it was.
+fn strip_nullable<'a>(schema: &'a Value, defs: &'a Map<String, Value>) -> &'a Value {
+    let Some(arms) = schema.get("anyOf").and_then(Value::as_array) else {
+        return schema;
+    };
+    let non_null: Vec<&Value> = arms
+        .iter()
+        .filter(|arm| arm.get("type").and_then(Value::as_str) != Some("null"))
```
TurtIeSocks, line 56: `strip_nullable` recognises a null arm only as the string `"null"`. JSON Schema also allows `{"type": ["null"]}`, and that arm survives the filter, leaves `non_null` with more than one element, falls to `_ => schema`, and the `anyOf` is never unwrapped, so the field stays `Opaque`.

Worth a look at whether schemars can emit that form for shep's own config before treating it as a gap: if it cannot, the current narrowness is correct and the doc could say so.

crates/shep-cli/src/lookout/pane/edit.rs
```diff
@@ -544,7 +581,7 @@ mod tests {
             pane.cycle();
             pane.begin_typing();
             pane.type_char('x');
-            pane.apply_typing();
+            assert_eq!(pane.apply_typing(), None, "the field takes what was typed");
```
TurtIeSocks, line 584: `apply_typing` returns `Option<Refusal>`, so `None` here means validation did not refuse. The message says the field takes what was typed, and the next line asserts the edit never reached the set, which is the test's actual point.

Something like "validation passes, the lock still blocks filing" matches the pair and keeps the reason (the lock, not validation) in front of the next reader.

crates/shep-cli/src/lookout/field/init.rs
TurtIeSocks: Pre-existing on main at `field.rs:353` and `:366`, byte for byte. Misfiled as a review comment on a split that only moved it. Filed as #593, resolving here.

crates/shep-cli/src/lookout/field/mod.rs
TurtIeSocks: Pre-existing on main at `field.rs:404`, byte for byte. Filed as #591, resolving here.

crates/shep-cli/src/lookout/field/schema.rs
TurtIeSocks: Pre-existing on main at `field.rs:315`, and already open as #561. Resolving here.

crates/shep-cli/src/lookout/field/schema.rs
TurtIeSocks: Pre-existing on main. Cosmetic, so not worth its own issue. Resolving here.

crates/shep-cli/src/lookout/field/schema.rs
TurtIeSocks: Pre-existing on main at `field.rs:300`. Cosmetic, so not worth its own issue. Resolving here.

crates/shep-cli/src/lookout/field/schema.rs
TurtIeSocks: Pre-existing on main at `field.rs:269`, byte for byte. Filed as #592, resolving here.

crates/shep-cli/src/lookout/field/fixtures.rs
```diff
@@ -0,0 +1,20 @@
+//! Test fixtures shared by this module's four halves.
+//!
+//! [`real_field_set`] reads the Flockfile schema off disk, which every
+//! half has a claim to make about: `bounds` that a range is published,
+//! `schema` that a grammar is marked, `init` that a neighbour names a
+//! real field. One copy, so a change to the schema moves one fixture.
+
+use super::FieldSet;
+
+pub(super) fn props(v: serde_json::Value) -> serde_json::Map<String, serde_json::Value> {
```
TurtIeSocks, line 10: `real_field_set` below carries a doc and this does not, in a file whose whole job is to be read by four other modules. It also panics when `v` is not an object, which the signature does not say.

It was a test-module helper on main, so no doc was owed. It is `pub(super)` here.

```suggestion
/// The `properties` map of `v`, which must be a JSON object.
pub(super) fn props(v: serde_json::Value) -> serde_json::Map<String, serde_json::Value> {
```

TurtIeSocks:
@coderabbitai review

crates/shep-cli/src/lookout/pane/edit.rs
TurtIeSocks: Fixed in d7624301. The message names the lock now. Checked the other seven sites of the same assertion: the original wording is accurate on each, so this was one instance.

crates/shep-cli/src/lookout/field/fixtures.rs
TurtIeSocks: Fixed in d7624301, took the suggestion.

TurtIeSocks:
@coderabbitai review

TurtIeSocks:
@coderabbitai review

TurtIeSocks:
@coderabbitai review

crates/shep-cli/src/lookout/validation.rs
```diff
@@ -61,6 +68,28 @@ pub const MEMORY_FORMS: &[Form] = &[
...
+    },
+    Form {
+        text: "a lowercase unit",
+        examples: &["512m", "2g"],
+    },
+    Form {
+        text: "a fraction",
+        examples: &["1.5G"],
+    },
+];
```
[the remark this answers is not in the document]
TurtIeSocks: Added in 381b0934. The assertion is `<=`, so all three entries pass as written. Checked the row is live by naming three forms against two examples, which fails with `MEMORY_REFUSALS: text "..." names 3 forms but examples has only 2`.

